PR #54, opened 2014-10-27
Add migration for existing model

danubekas:
Hi,

I had an existing model that i wanted to add devise_token_auth to but there is only a migration for creating a new user model. I added a template for adding the relevant columns to an existing model. It is working but could use some refactoring :)


wrburgess:
I'll vote for having this pulled. I am stuck with an existing Devise install, as well and the installation migration doesn't work.

lkkadiri:
@danubekas 

Im afraid you might have to rebase, looks like there could be some merge conflicts in it.

booleanbetrayal:
Going to have to close due to inactivity. I can re-open if you rebase / update for new changes (json support etc) @danubekas

PR #93, opened 2014-12-26
Make default Warden scope changeable

troggy:
Hi!

I have been using devise_token_auth successfully in my app until I need to change model name from User to Account. I reconfigured all the things as needed, but my application started throwing CookieOverflow exception:
````
ActionDispatch::Cookies::CookieOverflow (ActionDispatch::Cookies::CookieOverflow):
  actionpack (4.1.6) lib/action_dispatch/middleware/cookies.rb:529:in `[]='
  actionpack (4.1.6) lib/action_dispatch/middleware/session/cookie_store.rb:110:in `set_cookie'
````
In debugger I see that the whole Account object being put to session under ``warden.user.user.key`` which results to cookie overflow. If model is User than only user id is kept in session. I've noticed that gem always uses :user scope (which forms that ``warden.user.user.key`` name)  when talking to Devise and when I change scope name to :account error is gone and everything works as expected. So I decided to make scope configurable

Please note that I'm kind of new to Rails, so I may be tackling the problem from the wrong angle.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -38,7 +38,7 @@ module DeviseTokenAuth::Concerns::SetUserByToken
     user = uid && rc.find_by_uid(uid)
 
     if user && user.valid_token?(@token, @client_id)
-      sign_in(:user, user, store: false, bypass: true)
+      sign_in(DeviseTokenAuth.scope, user, store: false, bypass: true)
       return @resource = user
     else
       # zero all values previously set values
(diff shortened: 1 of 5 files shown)


lynndylanhurley:
Thanks @troggy - I'll review this ASAP!

mkhatib:
I am seeing the CookieOverflow problem with Twitter Oauth

nicolas-besnard:
I things this configuration doesn't allow you to have multiple model such as `User` and `Peon` right ?

booleanbetrayal:
Closing due to inactivity

PR #95, opened 2014-12-29
Make expiry key symbolized

jartek:
I'm currently getting a null expiry header,

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L163

Changing it to a symbol fixed it, so i assumed that the same issue would occur for all places where we used ```self.tokens[client_id]['<key>']```.

Edit: Changing it everywhere broke the suite, so i changed it to just the expiry header.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -160,7 +160,7 @@ module DeviseTokenAuth::Concerns::User
 
     # client may use expiry to prevent validation request if expired
     # must be cast as string or headers will break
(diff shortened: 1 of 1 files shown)


lynndylanhurley:
Thanks @jartek - I'll review this ASAP!

lynndylanhurley:
@jartek - I'm not sure why, but sometimes the `tokens` hash uses string keys, and other times it uses symbol keys.

I think I've fixed most of the issues in version `0.1.31.beta8`. Please pull it down and let me know if you run into any trouble.

PR #113, opened 2015-01-11
document GET for /validate_token

lukaselmer:
See https://github.com/lynndylanhurley/devise_token_auth/blob/master/lib/devise_token_auth/rails/routes.rb#L48


lynndylanhurley:
Thanks!!

PR #115, opened 2015-01-15
Use I18n for error messages

lukaselmer:


felipegs:
There are anothers messages that should be placed in internationalization file:
https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/passwords_controller.rb

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/registrations_controller.rb

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/sessions_controller.rb

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/token_validations_controller.rb

lukaselmer:
Indeed, thanks for pointing that out. I'll do it.

lynndylanhurley:
Thanks @lukaselmer! Let me know when those other files are updated and I'll merge ASAP!

Tiagoperes:
Will the messages in the files pointed by @felipegs be transfered to the internationalization file? If not, is there another way to translate them based in the user's language?

n5ken:
+1, I need i18n, it's great if this gem is able to support this function.

oniofchaos:
+1 this would be very useful to have completed.

booleanbetrayal:
Closing in favor of #283

PR #125, opened 2015-01-31
Check if confirmable is active before skipping confirmation

nicolas-besnard:


lynndylanhurley:
Thanks @nicolas-besnard!!

PR #143, opened 2015-02-09
Actual header key uses dashes, not underscores.

ragaskar:
Was using this project today and am not planning on using the angular bits. While writing request specs for the behavior I noticed that all header keys use dashes, not underscores as the README implies. I probably missed a few, but hopefully this is good enough for folks who are using the README as a quickstart to figure out what header values they care about.


lynndylanhurley:
Thanks @ragaskar!!!

Issue #176, opened 2015-03-09
confirm_success_url shouldn't be a required param


aesnyder:
I'm probably missing the point but...

It seems odd that devise_token_auth requires me to pass a confirm_success_url with the creation of the user. 

While it is nice to be able to control this at the request level, I would think that it's default behavior should be to use a default confirm_success_url defined somewhere in the rails layer.

I'm happy to open a pr, but wanted to get some weigh-in on where this makes the most sense to be defined?

nicolas-besnard:
I don't understand why this is set on the front side. I think the best solution should be to determine the URL on the configuration file of this gem.

aesnyder:
I agree I plan to add a commit to have a default confirm success url in the
rails config.

I can add.

lynndylanhurley:
The `confirm_success_url` param is used to support multiple client applications. If the param isn't required, then the API has no way of knowing where to redirect after email confirmation.

I'm open to allowing an optional default `confirm_success_url` to be set on the API. Feel free to send a PR if you have time.

lynndylanhurley:
I just pushed a fix for this in version `0.1.32.beta8`. You can now set defaults for the email confirmation and password reset redirects. See [the config section of the readme](https://github.com/lynndylanhurley/devise_token_auth#initializer-settings) for more info.

rohitahuja:
@lynndylanhurley I am currently trying to set default_confirm_success_url in devise_token_auth.rb, as so:

> config.default_confirm_success_url = "confirmed"

but I'm getting the error

> undefined method `default_confirm_success_url=' for DeviseTokenAuth:Module (NoMethodError)

PR #177, opened 2015-03-10
make confirm_success_url paramater optional

aesnyder:
As described in  #176


lynndylanhurley:
Hi @aesnyder - thanks for taking the time to send this PR.

I have a question about this change. When a user clicks the confirmation link in an email after registering, what would happen after a successful confirmation if the there is no redirect url defined?

lynndylanhurley:
This has been superseded by the new [`default_confirm_success_url` param](https://github.com/lynndylanhurley/devise_token_auth#configuration-cont).

PR #179, opened 2015-03-12
Ignore 'extra' in Twitter auth response to avoid CookieOverflow. Fixes #145.

### LINKED ISSUE #145: Twitter OAuth always throughs CookieOverflow
I've tried to fix this up and a lot of what I read online suggests to either not use cookies for the session storage or to remove some attributes from the auth response from twitter which has extra parameter which seems to be too big to store.

tbloncar:
This addresses #145.

### DIFF
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -11,8 +11,9 @@ module DeviseTokenAuth
       devise_mapping = request.env['omniauth.params']['resource_class'].underscore.to_sym
       redirect_route = "/#{Devise.mappings[devise_mapping].as_json["path"]}/#{params[:provider]}/callback"
 
-      # preserve omniauth info for success route
-      session['dta.omniauth.auth'] = request.env['omniauth.auth']
+      # preserve omniauth info for success route. ignore 'extra' in twitter
+      # auth response to avoid CookieOverflow.
+      session['dta.omniauth.auth'] = request.env['omniauth.auth'].except('extra')
       session['dta.omniauth.params'] = request.env['omniauth.params']
 
       redirect_to redirect_route


lynndylanhurley:
Thx @tbloncar!!

lynndylanhurley:
You're now [an official contributor](https://github.com/lynndylanhurley/devise_token_auth#callouts)!

tbloncar:
Thanks, @lynndylanhurley.

lynndylanhurley:
I just pushed version `0.1.32.beta7`. Please pull it down and confirm that it works ok.

nicolas-besnard:
What does it change to remove the 'extra' parameter ?

lynndylanhurley:
The schema description is [here](https://github.com/intridea/omniauth/wiki/Auth-Hash-Schema). I'm not sure exactly - I think it varies by provider.

nicolas-besnard:
I'm not sure this is the right think to do. I'm using this `extra` parameters on my projects.

tbloncar:
@nicolas-besnard Your point makes sense to me; I think the issue here is that the default behavior for Twitter auth (before this change) was a raise of `ActionDispatch::Cookies::CookieOverflow`. While this change ameliorates that issue, I see that it may result in missing some of the data you mentioned.

One alternative to this solution is making some mention of [using an alternate session store](http://stackoverflow.com/a/7117725) (like ActiveRecord store) in the README. This would be "as needed," of course. Another option might be using `except('extra')` for only the Twitter callback (for now).

c0mrade:
Yeah I got the same `CookieOverflow` from google too, and I'm using a database as a data store. It all depends what info are you looking to get. It may or may not resolve in the issue. Yeah doesn't sound right to restrict people who want to get that extra info (I myself don't use or need it), just thinking out loud.

nicolas-besnard:
As this is an API, we don't need Cookie or Session. I just think the
implementation of Omniauth is not valid on this gem. I'll dig this this
weekend.

c0mrade:
Yes that was my initial thought but I had some errors and had to deliver something, so I ended up adding session middleware so the code works, thanks for reminding me I will need to remove that.

medaglia:
I too am using omniauth_facebook and require the data passed via `extra` (age_range, gender, etc). I wasn't able to override redirect_callbacks (whereas i have no problem overriding omniauth_failure).  Does anyone have any any good suggestions? Right now I'm torn between...
1. branching omniauth_facebook to pass the data through `info` instead of `extra`
2. branching devise_token_auth to keep `extra`

PR #198, opened 2015-03-26
Refactor omniauth_success to ease multiple login methods

dcadenas:
I extracted the section that identifies the user logged in through oauth into a method that can be easily overriden so that someone that logged in through email/password can also log in through oauth.

Example:

``` ruby
module Overrides
  class OmniauthCallbacksController < DeviseTokenAuth::OmniauthCallbacksController

    def resource_identification_hash
      { email: auth_hash['info']['email'] }
    end
  end
end
```


justinsoong:
@dcadenas +1

nbrustein:
@nicolas-besnard can you be more specific about what the change would look like in OmniauthCallbackController?  How do we use that policy to find the resource?

Issue #196, opened 2015-03-24
I have a rails backend rendered app (erb). Can I switch to devise token auth?


casertap:
I would like to know if it is possible to use token auth on a simple rails application (no api, no angular, just simple .erb views)?
I have some trouble implementing it.

nicolas-besnard:
Why do you want to use this gem instead of Devise ? What are your troubles ?

casertap:
Actually, I have a coupled version of rails-angular.
It looks a bit like this: https://github.com/colindensem/demo-rails-angularjs
But some parts of the website are rails backend rendered (erb) so I was rendering if I can have devise token auth so I will have less troubles latter when I totally switch to a decoupled angular app (in public dir).

nickL:
@casertap See my comment in #120 -- Does that help?

jotolo:
I was working in a similar case where I needed to use both, devise+devise_token_auth in a spree application. I was trying to implement authentication for spree_api and migrating the authentication of spree system(Full Stack App) to devise.
I found a way to do it and it's very simple.It's not a common thing in rails but it works great. The idea is to have a main application_controller.rb in your application and particular application_controllers for each section. In my case one for admin system in spree and the other one for API.
```
#controllers/api/v1/application_controller.rb
module Api
  module V1
    class ApplicationController < ApplicationController
      skip_before_action :verify_authenticity_token
      include DeviseTokenAuth::Concerns::SetUserByToken
    end
  end
end
```
In the API controller I included the regular DeviseTokenAuth concern to manage auth in the app that controls this application_controller.rb. Note that this controller inherits from the main application_controller.
```
#controllers/admin/application_controller.rb
  module Admin
    class ApplicationController < ApplicationController
      before_action :authenticate_user!
    end
  end
```
This controller is, well, natural behavior of devise. I just needed to add the before action method.
We just need the base application_controller ;)
```
class ApplicationController < ActionController::Base
  protect_from_forgery with: :exception, if: :verify_api

  def verify_api
    params[:controller].split('/')[0] != 'devise_token_auth'
  end
end
```
Our main application_controller.rb looks like this. You need to keep the CSRF token protection but you can avoid it when an API request arrive.
The idea is to redirect each flow in your application. the ones that belongs to your full-stack app and the other ones that belongs to your API. Well, in this case to my API :smile: 
[...]

Issue #168, opened 2015-03-03
Sign_in / Sign_up via token_auth and via session


patricks:
Hello,
I have setup devise_token_auth for api / json communication and everything works nicely. I can sign_up / sign_in ... and get json responses.
Now I want some features to accessible via a web interface. How do I setup a sign_up / sign_in via this gem? Or is this possbile? 
I tried to generate devise views (like it did on other projects, where i only use devise) via rails generate devise:views. Put this doesn't work. Is this even possible with this gem? Or maybe there is an alternative gem?

nicolas-besnard:
You can use [this library for AngularJS](https://github.com/lynndylanhurley/ng-token-auth) which was created by the creator of this gem.

nickL:
I think you would need to use normal Devise for your web routes and DeviseTokenAth for your api/json communication. See #120, i believe there's a similar need there?

zachfeldman:
Hi there @patricks   ,

In an effort to cleanup this project and prioritize a bit, we're marking issues that haven't had any activity in a while with a "close-in-7-days" label. If we don't hear from you in about a week, we'll be closing this issue. Obviously feel free to re-open it at any time if it's the right time or this was done in error! 

If you are still having the issue (especially if it's a bug report) please refer to our new [Issue Template](https://github.com/lynndylanhurley/devise_token_auth/blob/master/.github/ISSUE_TEMPLATE.md) to provide some more details to help us solve it.

Hope all is well.

Issue #120, opened 2015-01-26
Having both devise and devise_token_auth


casertap:
I have a lot of trouble to set up my application with both devise and devise_token_auth.
We are slowly migrating from a rails server-side rendering to an angularjs application.
The problem is that I want to have a way to log in and have part of my website with angular and part without angular (yet). 
It seems that with devise_token_auth the current_user is not statefull, that is a problem for me because I need to log in ones and then I just need to check if current_user exists.
At least I need to do that until the migration is complete and I can totally switch to a token auth.
Do you have an idea on how to solve my problem?

lynndylanhurley:
I'm having a hard time understanding this issue. Can you please elaborate? For example, what do you mean by "stateful"?

Also, have you read [this section](https://github.com/lynndylanhurley/devise_token_auth#can-i-use-this-gem-alongside-standard-devise) of the README?

casertap:
Ok I will try to be more clear.
I actually have an application that use Devise to loggin/signin etc...
So when I currently loggin, a new current_user is created and then I keep this user until he logout.

We are migrating page after page to angularJS so I was wondering if it is possible to have devise_token_auth on pages with angular and my standard Devise for my old pages.
The thing is that I only want an unique way to loggin.

nickL:
@casertap Yes, you can have normal (cookie) based devise for your legacy pages and, as you migrate to angular, use DeviseTokenAuth for your new angular pages. The way I'm handling it now is having the `DeviseTokenAuth::Concerns::SetUserByToken` check for a `warden.user` in session before checking the token headers.  For old pages -- the user is authenticated via devise. For new angular pages the user is authenticated via DeviseTokenAuth (via ng-token-auth).

lynndylanhurley:
> The way I'm handling it now is having the DeviseTokenAuth::Concerns::SetUserByToken check for a warden.user in session before checking the token headers.

@nickL - that sounds like a good strategy. Should we incorporate this behavior into the gem?

nickL:
@lynndylanhurley: Yeah, I think it may be useful. Here's a quick snippet. Let me know if I'm on the right track and I'll clean it up more and submit a PR:

```ruby

  def set_user_by_token(mapping=nil)
    # determine target authentication class
    rc = resource_class(mapping)

    # no default user defined
    return unless rc

    # user has already been found and authenticated
    return @resource if @resource and @resource.class == rc

    # parse header for values necessary for authentication
[...]

PR #200, opened 2015-03-30
Authenticating an existing Warden/Devise User

nickL:
Yo!

Alright, so I *think* this will help with #120, #168,  & #196;  In the event an app uses devise for normal session authentication and would like to also authenticate users via `DeviseTokenAuth` then I think this will help support both strategies and ease the transition.   Small bit of interstitial code to check for the user in warden and, if they're missing authentication details, create a token for them and pass them through. Otherwise, the normal token auth strategy kicks in.

**I had a couple of questions that are worth mentioning before merging this in, if it feels like its on the right approach:**

1. I added a test in `DemoUserController` integration test.  Does that feel like a good place for it?

2. I unfortunately can't use this commit in our own project because we're still on Rails 4.1, lol. Still wanted to contribute! We definitely have plans to upgrade soon, but is there anyway to support this in any earlier version?  I suspect no, which is perfectly fine! Just more gas on the fire to get us to upgrade.  I was just curious if there was a way to allow any Rails 4.1 apps to use it.

3. Anything I missed? Let me know. I'll do the work. You're busy enough.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -21,18 +21,25 @@ module DeviseTokenAuth::Concerns::SetUserByToken
     # no default user defined
     return unless rc
 
-    # user has already been found and authenticated
-    return @resource if @resource and @resource.class == rc
-
     # parse header for values necessary for authentication
     uid        = request.headers['uid'] || params['uid']
     @token     = request.headers['access-token'] || params['access-token']
     @client_id = request.headers['client'] || params['client']
 
-    return false unless @token
-
     # client_id isn't required, set to 'default' if absent
     @client_id ||= 'default'
+
+    # check for an existing user, authenticated via warden/devise
+    devise_warden_user =  warden.user(rc.to_s.underscore.to_sym)
+    if devise_warden_user && devise_warden_user.tokens[@client_id].nil?
+      @resource = devise_warden_user
+      @resource.create_new_auth_token
+    end
+
+    # user has already been found and authenticated
+    return @resource if @resource and @resource.class == rc
+
+    return false unless @token
 
     # mitigate timing attacks by finding by uid instead of auth token
     user = uid && rc.find_by_uid(uid)
--- a/test/controllers/demo_user_controller_test.rb
+++ b/test/controllers/demo_user_controller_test.rb
@@ -7,6 +7,7 @@ require 'test_helper'
 #  was the appropriate message delivered in the json payload?
 
 class DemoUserControllerTest < ActionDispatch::IntegrationTest
+  include Warden::Test::Helpers
   describe DemoUserController do
     describe "Token access" do
       before do
@@ -258,5 +259,57 @@ class DemoUserControllerTest < ActionDispatch::IntegrationTest
         end
       end
     end
+
+    describe 'Existing Warden authentication' do
+      before do
+        @resource = users(:confirmed_email_user)
+        @resource.skip_confirmation!
+        @resource.save!
+        login_as( @resource, :scope => :user)
+
+        # no auth headers sent, testing that warden authenticates correctly.
+        get '/demo/members_only', {}, nil
+
+        @resp_token       = response.headers['access-token']
+        @resp_client_id   = response.headers['client']
+        @resp_expiry      = response.headers['expiry']
+        @resp_uid         = response.headers['uid']
+      end
+
+      describe 'devise mappings' do
+        it 'should define current_user' do
+          assert_equal @resource, @controller.current_user
+        end
+
+        it 'should define user_signed_in?' do
+          assert @controller.user_signed_in?
+        end
+
+        it 'should not define current_mang' do
+          refute_equal @resource, @controller.current_mang
+        end
+      end
+
+      it 'should return success status' do
+        assert_equal 200, response.status
+      end
+
+      it 'should receive new token after successful request' do
+        assert @resp_token
+      end
+
+      it 'should set the token expiry in the auth header' do
+        assert @resp_expiry
+      end
+
+      it 'should return the client id in the auth header' do
+        assert @resp_client_id
+      end
+
+      it "should return the user's uid in the auth header" do
+        assert @resp_uid
+      end
+    end
+
   end
 end


lynndylanhurley:
Fantastic PR @nickL. You just made my day with these tests.

> I added a test in DemoUserController integration test. Does that feel like a good place for it?

`DemoUserController` is the best place - that's where I'm currently doing all of the integration tests.

> is there anyway to support this in any earlier version?

I have been getting asked this a lot. I don't remember exactly what the incompatibility is, but I do remember that I had to change some things for Rails 4.2. I'll downgrade to Rails 4.1 and see if the tests still pass.

Is there a way to have the test suite run against multiple versions of Rails?

casertap:
Thanks nickL for your work.
I need to use this code to be able to login with devise_token_auth and still have some old erb portal accessible for logged admin users.

I encounter two problems with the solution:
1) the conditional in set_user_by_token
````
````
if never true because I am logged in with devise_token_auth and so my token is set. 
I think, you should remove the second part of the conditional and check if this does not introduce some security vulnerability.

2) when I visit my old pages, this code in  in set_user_by_token
````
@token     = request.headers['access-token'] || params['access-token']
`````
@token will be nil but it is still included in the header (as access-token)
This make crash rake/lib/rack/handler/webrick.rb:100
```
headers.each { |k, vs|
            next if k.downcase == "rack.hijack"
            if k.downcase == "set-cookie"
              res.cookies.concat vs.split("\n")
            else
              # Since WEBrick won't accept repeated headers,
              # merge the values per RFC 1945 section 4.2.
              res[k] = vs.split("\n").join(", ")
            end
          }
````
because he is trying the do nil.split("\n").join(", ") because access-token is nil.
I think, you should remove the access-token header if this one is nil.

Let me know. Thanks a lot.

**Comment by casertap** [comment]
Actually there are many other problems:
3)
If I log with a user 1 and then logout and log in with another user 2.
because the logout of warden was not called you have the user 2 that can connect using its password and access directly the account of the user 1.

4)
If you have an user with a tokens empty.
If this option is setup (that is my case in development mode)
````
  config.change_headers_on_each_request = true
end
````
than you will get this stack
````
NoMethodError - undefined method `[]' for nil:NilClass:
   () Users/pierrecaserta/.rvm/gems/ruby-2.2.1/bundler/gems/devise_token_auth-a561a9b74bd8/app/models/devise_token_auth/concerns/user.rb:169:in `build_auth_header'
   () Users/pierrecaserta/.rvm/gems/ruby-2.2.1/bundler/gems/devise_token_auth-a561a9b74bd8/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:73:in `block in update_auth_header'
  activerecord (4.2.1) lib/active_record/locking/pessimistic.rb:72:in `block in with_lock'
  activerecord (4.2.1) lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
  activerecord (4.2.1) lib/active_record/connection_adapters/abstract/transaction.rb:188:in `within_new_transaction'
  activerecord (4.2.1) lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
  activerecord (4.2.1) lib/active_record/transactions.rb:220:in `transaction'
  activerecord (4.2.1) lib/active_record/transactions.rb:277:in `transaction'
  activerecord (4.2.1) lib/active_record/locking/pessimistic.rb:70:in `with_lock'
   () Users/pierrecaserta/.rvm/gems/ruby-2.2.1/bundler/gems/devise_token_auth-a561a9b74bd8/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:64:in `update_auth_header'
````

the code responsible for this error is :
````
 def update_auth_header
    # cannot save object if model has invalid params
    return unless @resource and @resource.valid? and @client_id
    # Lock the user record during any auth_header updates to ensure
    # we don't have write contention from multiple threads
    @resource.with_lock do
      # determine batch request status after request processing, in case
      # another processes has updated it during that processing
      @is_batch_request = is_batch_request?(@resource, @client_id)
      auth_header = {}
      if not DeviseTokenAuth.change_headers_on_each_request
        auth_header = @resource.build_auth_header(@token, @client_id)
````
You are trying to build_auth_header but you are actually still connected with warden so @token is nil

**Comment by nickL** [comment]
Hi @casertap, I'm sorry you're running into such trouble! I'm not familiar with the error you describe  re: warden/devise_token_auth and I'm having trouble reproducing. Though I'll do my best to help if I can.  To get me started can you do either of the following to help me reproduce?

1) Write a simple test-unit/rspec test that demonstrates the above error?
2) Link me to a fork with steps to reproduce?
3) Give a bit more detail (step by step) of how you're **logging in** and **logging out** your two different users. I'm a bit unclear how to reproduce this is a blank app.

1 or 2 would be preferred, it'll help me try and reproduce this in the way you described. Happy to help if I can.

Thanks!

**Comment by booleanbetrayal** [comment]
there's a few PRs that have just gone in, so any work you might make against this issue, be sure to do so on a fresh pull of master. for instance, `reset_session` is now being called on session destroy.

**Comment by nbrustein** [comment]
@casertap @nickL  I'm running to what sounds like kind of a similar issue to @casertap, but I don't understand how.

We set the `storage` config value to `'localStorage'`.  We also have rails sessions turned on, only because we need them turned on to support omniauth.  Unlike @casertap, the only way to login to our site is with devise token auth.

Yet, once every few days, we run into a state where a request comes in with no tokens, and the user is still found by:  `warden.user(rc.to_s.underscore.to_sym)`  As far as I understand, this means that the user does not have a token stored in localStorage, but they do have a session id stored in a cookie.  

It's a long story about how we noticed this.  I don't think it's causing us any direct problems, but it makes me nervous that I don't really understand what "logged in" really means on my site right now.

I can reproduce this issue by logging in and then deleting my local storage in the Chrome dev tools.  This makes sense because, in that case, I still have a session_id stored in a cookie, so the next request sends up a session_id but no token.  However, I do not believe that my real uses are out there deleting local storage values but not deleting cookies.

Before #251, this would happen after an explicit logout, since the logout removed the token from the auth headers but did not remove the session_id from the cookie, but #251 solved that.  

So now I don't know how it's happening.  Do you have any other explanation of how someone could get into this state where they send up no token but still get logged in via warden?

**Comment by booleanbetrayal** [comment]
@nickL / @casertap - Due to the issues resulting from this warden implementation, I'm inclined to make this warden lookup configurable under some sort of `support_legacy_authentication` flag of some sort, unless you can indicate why we would want to support this under all use cases. Please provide any insight you might have!

**Comment by ianchen06** [comment]
@nbrustein  @booleanbetrayal - I'm running into the same issue as @casertap , 
```
3)
If I log with a user 1 and then logout and log in with another user 2.
because the logout of warden was not called you have the user 2 that can connect using its password and access directly the account of the user 1.
```
Please advise if there are workarounds or patches for this issue. We need the sessions enabled for omniauth login.

Thanks a lot for your help and a great gem.

**Comment by nickL** [comment]
@booleanbetrayal , @casertap :  Sorry!!  I missed the notifications on this thread and just saw your replies. My mistake for not replying sooner. 

@booleanbetrayal : I like the idea of making this configurable  I don't think it's a use-case for everyone as  soome implementations may not use warden this way. I'd support a configurable toggle if that would be sufficient. I don't mind helping with this if you want to talk through it.

**Comment by lynndylanhurley** [comment]
@nickL - this commit seems to be causing problems with logout (see #375). I just pushed a change to address it (see https://github.com/lynndylanhurley/devise_token_auth/commit/c80cfce88c4fbfce163dde4a7527d0c13f30d77f). It didn't break any tests, but can you please pull down version `0.1.37-beta1` and see if everything still works ok with your setup?

**Comment by booleanbetrayal** [comment]
@lynndylanhurley - we're seeing the `sign_out(@resource)` call clear out the user record prior to certain flows in our controller logic, such as during [CanCan](https://github.com/ryanb/cancan) validations. Should we be deferring this somehow, or clearing the user in a way that only effects the session store for Devise?

**Comment by lynndylanhurley** [comment]
@booleanbetrayal - That's what I was afraid of.

This line seems to be the real problem:

~~~ruby
devise_warden_user =  warden.user(rc.to_s.underscore.to_sym)
~~~

I think it was intended to pull the data from an existing devise session (standard devise, not devise_token_auth), but it seems to be also instructing the browser to set a cookie containing the user's devise session info.

This is problematic because the API shouldn't deal with cookies at all (that is the purpose of token based authentication). But I understand why this functionality is useful.

So the question is, can we accomplish what this line of code is doing (identifying the user from the standard devise session) without actually setting the cookie?

**Comment by booleanbetrayal** [comment]
i dunno ... i was toying with ripping that support out entirely or at the very least make it opt-in. here's how we're dealing with it today. it's a hacky hodge-podge of `reset_session` calls:

## application_controller.rb

```ruby
    rescue_from NoMethodError, with: :catch_devise_error
    def capture_devise_no_method_error(exception)
        if exception.message == "undefined method `[]' for nil:NilClass" && 
            exception.backtrace[0].match(/build_auth_header/)
            # log the exception, etc etc ...
            reset_session
            return true
        else
            return false
        end
    end
```

## home_controller.rb

```ruby
before_action :reset_session # NOTE: this is a little misleading. the session is only used briefly and only during omniauth (?).
```

## devise_token_auth.rb (monkey-patch)

```ruby
require 'devise_token_auth/sessions_controller'
module DeviseTokenAuth
    class SessionsController < DeviseTokenAuth::ApplicationController

        after_action :reset_session, :only => [:destroy]

    end
end
```

lynndylanhurley:
This would be easier if we had a failing test that we could solve against. I'll see what I can come up with tonight.

lynndylanhurley:
@booleanbetrayal - can you please try version `0.1.37.beta3`? I think I was able to get around the issue without calling `sign_out`.

booleanbetrayal:
@lynndylanhurley - I'm still experiencing the same issue as when you were explicitly calling `sign_out`.

Can is wired up via a `before_filter` and by the time I'm seeing it register enter play, `current_user` is already `nil`

https://github.com/ryanb/cancan/blob/master/lib/cancan/controller_additions.rb#L357

```ruby
 "/Users/brent/.rvm/gems/ruby-2.2.2@back_royal/gems/cancan-1.6.10/lib/cancan/controller_additions.rb:357:in `new'",
 "/Users/brent/.rvm/gems/ruby-2.2.2@back_royal/gems/cancan-1.6.10/lib/cancan/controller_additions.rb:357:in `current_ability'",
 "/Users/brent/.rvm/gems/ruby-2.2.2@back_royal/gems/cancan-1.6.10/lib/cancan/controller_additions.rb:338:in `authorize!'",
 "/Users/brent/.rvm/gems/ruby-2.2.2@back_royal/gems/cancan-1.6.10/lib/cancan/controller_resource.rb:41:in `authorize_resource'",
 "/Users/brent/.rvm/gems/ruby-2.2.2@back_royal/gems/cancan-1.6.10/lib/cancan/controller_resource.rb:10:in `block in add_before_filter'",
...
```

lynndylanhurley:
@booleanbetrayal - `current_user` should be available after the `set_user_by_token` `before_filter` is called. Is it possible that the cancan filters are being run before the `set_user_by_token` filter?

booleanbetrayal:
I'm always seeing `set_user_by_token` getting called prior to the CanCan initialization. maybe something internal to Devise? I'm wondering if we shouldn't just make this an opt-in setting and punt for now. Think it'd result in improved stability for most users.

lynndylanhurley:
> I'm wondering if we shouldn't just make this an opt-in setting and punt for now. Think it'd result in improved stability for most users.

Agreed. I was at this all night last night, and I'm still not sure how to write a test for this :dizzy_face:

I'm speculating here, but it looks like Devise/warden is getting the browser to set a cookie that correlates to some kind of warden session.

That session is actually destroyed on the server when the `sessions#destroy` route is hit. But the cookie still persists on the browser, and the presence of that cookie seems to revive the session somehow (may be related to this issue: https://github.com/plataformatec/devise/issues/3031).

The only I could find around this was to prevent the creation of that cookie. It would be better to have the browser terminate the cookie upon logout. I'll keep looking for a way to do that.

booleanbetrayal:
@lynndylanhurley - going to pick this up (configuration flag / new default) unless you've already started it

lynndylanhurley:
@booleanbetrayal - please do. 

I'll keep trying to figure out why the session persists after logout, and if there's anything that can be done to prevent it.

booleanbetrayal:
@lynndylanhurley - maybe take a quick peek at #428 and let me know if you see anything

booleanbetrayal:
This is now configurable as per #428 - defaulting to disabled

matisoffn:
I keep getting this error seems like it comes from this PR on this line: `devise_warden_user =  warden.user(rc.to_s.underscore.to_sym)`

I'm getting this:

```
CasesControllerTest#test_should_get_index:
NoMethodError: undefined method `user' for nil:NilClass
    devise_token_auth (0.1.36) app/controllers/devise_token_auth/concerns/set_user_by_token.rb:35:in `set_user_by_token'
```

Issue #203, opened 2015-04-05
500 error returned when no data is POSTed to registration controller


mchavarriagam:
Hello,

When posting an empty body to the Registration controller, a 500 error is returned:

    NoMethodError in DeviseTokenAuth::RegistrationsController#create
    undefined method `downcase' for nil:NilClass
    Extracted source (around line #12):
	
      # honor devise configuration for case_insensitive_keys
      if resource_class.case_insensitive_keys.include?(:email)
        -> @resource.email = sign_up_params[:email].downcase
      else
        @resource.email = sign_up_params[:email]
      end

I'm currently working on an API which uses DTA so I have no way of ensuring that the client *will* send proper sign up data in POST body. I can do a PR for this but I'd like to make sure I'm going the right way about it:

1. Adding before_filter on registrations_controller (validate_sign_up_body perhaps? only for post)
2. validate_sign_up_body returns 422 unprocessable entity + json errors: 'Please post proper sign up data in request body.' if sign_up_params is empty.

Does that sound good?

Cheers.

mchavarriagam:
Thank you!

PR #204, opened 2015-04-05
Return 422 (was 500) when empty body for sign up and account update

mchavarriagam:
Hello.

As mentioned on issue #203, currently DTA returns 500 error when an empty POST or PUT is made on the registration controller. For those using DTA for API authentication without control over REST clients, this may be an issue. I went ahead and made some changes. I think I can do some refactoring on the tests and also on the before filters, but wanted to check if the idea is welcomed before moving forward.

Thanks!

### DIFF
--- a/app/controllers/devise_token_auth/registrations_controller.rb
+++ b/app/controllers/devise_token_auth/registrations_controller.rb
@@ -1,6 +1,8 @@
 module DeviseTokenAuth
   class RegistrationsController < DeviseTokenAuth::ApplicationController
     before_filter :set_user_by_token, :only => [:destroy, :update]
+    before_filter :validate_sign_up_params, :only => :create
+    before_filter :validate_account_update_params, :only => :update
     skip_after_filter :update_auth_header, :only => [:create, :destroy]
 
     def create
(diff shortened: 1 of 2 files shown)


lynndylanhurley:
Thanks again @mchavarriagam, and thanks again for the tests!!

lynndylanhurley:
And congratulations, you've made it to the [callouts list](https://github.com/lynndylanhurley/devise_token_auth/blob/master/README.md#callouts).

alex88:
Shouldn't this be handled with validations? Like saying "Email shouldn't be blank"?

mchavarriagam:
@alex88 I'm not sure what the protocol should be here but my assumption was that multiple fields would have to be validated and I wasn't certain if the fields would be pre-defined and constant. Regardless of that, I didn't _feel_ it was appropriate for the gem to return 500 (server error) for something that was in reality a user error. This is just a long way of saying "yes it should probably be done with validations but I didn't know what to validate for so I just checked that the body wasn't empty" :) (that didn't turn out short either).

I'm open to suggestions on this and wouldn't mind working on the PR a bit more to make it better.

alex88:
@mchavarriagam maybe we should stay in line with devise https://github.com/plataformatec/devise/blob/master/lib/devise/models/validatable.rb#L29-L35 since "Please submit proper sign up data in request body" doesn't let the user know what's wrong with his submission neither it tells what to fix.

Devise in its controllers just tries to persist the entity, the checks are done on the validation side.

Maybe we can just use downcase only on non blank email params and let the validations do the work

PR #205, opened 2015-04-06
remove fragment sign ("#") from URLs without fragment

tomdov:
the "#" sign existed on URLs before the query sign "?" which made rails believe there are no params in the url (e.g. ventory.com#?param1=value1&param2=value2 was recognized as the request for ventory.com without params.).
removing the "#" sign was necessary in case there is no fragment in the url.

Thanks


lynndylanhurley:
Hi @tomdov - thank you for taking the time to send this PR!

This was originally a workaround for a bug in angular itself. I'll need to do a bunch of manual testing on this before I can merge. Hopefully I'll have time for this tomorrow or this weekend.

tomdov:
Hi,
have you checked this? can it be merged?
Thanks in advance and thanks for this gem!

booleanbetrayal:
Hit this in the wild the other day. LGTM!

PR #206, opened 2015-04-07
Ability to localize error message

lda:
Gem users must have ability to localize error messages.
So to localize error, simply put your message in locale.yml with key 
activerecord.errors.models.{{model}}.attributes.email.already_in_use


lynndylanhurley:
Hi @lda  -thanks for the PR! This looks good to merge. 

I've been meaning to refactor the error messages for better consistency (format, status codes, localization, etc.). Is this something that you would be interested in taking on?

PR #207, opened 2015-04-09
Super block yield

sgwilym:
I often find myself wanting to make small additions to the `RegistrationsController` (such as adding a role to a just-created User), but find overriding the entire controller a little bit overkill when I simply want to _add_ behaviour.

In vanilla Devise you can make a controller that inherits from `RegistrationsController`, override the method  and simply pass a block to `super` like this:

```ruby
def create
  super do |resource|
    resource.add_role :cool_user
  end
end
```

Which I think is ideal for small additions like these. This PR implements block yielding in `RegistrationsController` (and though it could easily be extended to the other controllers, I thought I'd see what you make of this before going the whole hog on this).

And sorry if the tests are a little gross looking — I'm used to Rspec!

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Please read the [issue reporting guidelines](#issue-reporting) before posting is
   * [Excluding Modules](#excluding-modules)
   * [Custom Controller Overrides](#custom-controller-overrides)
   * [Email Template Overrides](#email-template-overrides)
+  * [Passing blocks to Controllers](#passing-blocks-controllers)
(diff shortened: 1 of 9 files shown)


lynndylanhurley:
Fantastic PR @sgwilym! The tests look great!

| think this is definitely an improvement. Would you be interested in extending this behavior to the other controllers as well?

sgwilym:
Thanks! 😄 And sure thing. I'll scrape a little time together soon to get all the controllers all blocked up.

PR #209, opened 2015-04-10
Super block yield (all controllers)

sgwilym:
This expands on #207 by introducing block yielding to all of DeviseTokenAuth's controllers, allowing you to add behaviour to any controller's methods without having to reimplement the whole thing, like so:

```ruby
class Custom::RegistrationsController < DeviseTokenAuth::RegistrationsController

  def create
    super do |resource|
      resource.do_something(extra)
    end
  end

end
```

Please let me know if I've missed anything. I also didn't add block yielding to `OmniAuthCallbacksController`'s `redirect_callbacks`, as I'm not sure whether you'd _want_ to mess around in there. 

Have a good weekend!

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ Please read the [issue reporting guidelines](#issue-reporting) before posting is
   * [Using Multiple User Classes](#using-multiple-models)
   * [Excluding Modules](#excluding-modules)
   * [Custom Controller Overrides](#custom-controller-overrides)
-  * [Email Template Overrides](#email-template-overrides)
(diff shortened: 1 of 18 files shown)


booleanbetrayal:
@sgwilym - Quick scan of this branch looks good and will definitely help to clean up some code in the wild. Would you mind doing a quick rebase to flatten these commits into a single descriptive commit message? Thanks in advance!

sgwilym:
@booleanbetrayal Done and done.

booleanbetrayal:
Awesome ... thanks @sgwilym !

PR #220, opened 2015-04-17
Fix error when email missing from registration params

iangreenleaf:


booleanbetrayal:
LGTM! Thanks @iangreenleaf

PR #225, opened 2015-04-23
fixes password reset when not using confirmable

### LINKED ISSUE #224: pending_reconfirmation called when confirmable isn't used
Hey I'm not using :confirmable but when I try to send password reset via devise token auth I get an error for no method `pending_reconfrimation?` on user... I think that we need to check for confirmable before calling `pending_reconfirmation?` here: 

https://github.com/lynndylanhurley/devise_token_auth/blob/852912c93dcf7df8df4d7494fc5d57f10cd105bc/app/models/devise_token_auth/concerns/user.rb#L68

I'll open a pr if you think that's the right way to go.

aesnyder:
fixes #224

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -65,7 +65,7 @@ module DeviseTokenAuth::Concerns::User
       # fall back to "default" config name
       opts[:client_config] ||= "default"
 
-      if pending_reconfirmation?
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
LGTM!

aesnyder:
heh, I forgot I submitted this pr.. Thanks for the merge :tada:

booleanbetrayal:
trying to help @lynndylanhurley out a bit with this and ng-token-auth  =]

PR #242, opened 2015-05-13
Fix group authentication check

lumenlunae:
Previously still allowed access to routes with authentication. This
adds a test copied from the user controller test and returns a 401.


rstrobl:
Exactly what I need :) Thanks @lumenlunae  !!
Why is this not getting merged?

lumenlunae:
Their version is a few versions up now. I should update my own branch here to match master, and hopefully will have time in the next day or two too.

booleanbetrayal:
closed via #446

PR #251, opened 2015-05-22
fix(sessions controller): call reset_session on destroy

nbrustein:
I started using omniauth, which required me to enable rails sessions.  I noticed that the session was not getting killed when I logged out.  This should fix it.


booleanbetrayal:
:+1:

PR #252, opened 2015-05-22
feat(ominauth): support json-formatted values in omniauth callback. 

### LINKED ISSUE #221: omniauth_success.html.erb JSON bug
I realized that, in the current way omniauth_success.html.erb is written (copied below), the JSON returned gives booleans and numbers as strings, e.g. "false", rather than false or "5" rather than 5. This gave me some problems cause I use coffeescript, which only supports the === or !== equalities which means that "5" !== 5. Any way to fix this easily?

```
<% @resource.as_json.each do |attr, val| %>
  "<%= attr %>": "<%= val %>",
<% end %>

"auth_token": "<%= @token %>",
"message":    "deliverCredentials",
"client_id":  "<%= @client_id %>",
"expiry":     "<%= @expiry %>",
"config":     "<%= @config %>"
```

nbrustein:
Fixes #221

This is potentially a breaking change.  For example, nil values used to be rendered as an empty string, but will now render as undefined.

We needed it because we have some values in there that are arrays.

### DIFF
--- a/app/views/devise_token_auth/omniauth_success.html.erb
+++ b/app/views/devise_token_auth/omniauth_success.html.erb
@@ -1,5 +1,5 @@
 <% @resource.as_json.each do |attr, val| %>
-  "<%= attr %>": "<%= val %>",
+    "<%= attr %>": <%= val.to_json.html_safe %>,
 <% end %>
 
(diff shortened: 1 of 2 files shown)


booleanbetrayal:
:+1:

PR #254, opened 2015-05-27
fix(session#new): fix unhandled 500 when logging in with valid user and bad password

mathemagica:
When token rotation is disabled, and the devise @resource is accessed within a before_filter, incorrect login credentials for an existing account can fail ungracefully causing a 500 server error instead of the expected 401. This is because both the @resource and @client_id remain set, despite falling through the failure criteria in `set_user_by_token`. As these instance variables persist, they allow the after_filter update_auth_header to continue to execute unexpectedly, calling  `auth_header = @resource.build_auth_header(@token, @client_id)`. This then triggers a null-ref in build_auth_header.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -14,7 +14,6 @@ module DeviseTokenAuth::Concerns::SetUserByToken
 
   # user auth
   def set_user_by_token(mapping=nil)
-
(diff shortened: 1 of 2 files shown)


booleanbetrayal:
@lynndylanhurley - this was a crazy dive and probably not encountered by most people due to our disabling of the default token rotation and our use of `before_filter`s that check `current_user`. Please let us know if you have any questions!

barillax:
:+1: tricky bug!

PR #272, opened 2015-06-17
perf(token_is_current?): add simplistic cache to reduce overhead of redundant token checks during validation calls

booleanbetrayal:
Within `DeviseTokenAuth::Concerns::User#token_is_valid?` we're seeing the `BCrypt::Password.new(token_hash)` + string comparison (casting) operation itself take approximately 75ms on average when validating tokens, which makes it hard / impossible to reach our goal of < 200ms response times for request handling. 

This PR adds a very simplistic caching mechanism that will lookup the check value, saving unnecessary overhead on redundant token / hash checks. It can be iterated on further if a more robust caching implementation is needed / desired. 

This will primarily benefit environments where `change_headers_on_each_request` is disabled, but will still help requests that occur within the `batch_request_buffer_throttle` threshold.

Thanks to @nbrustein for the implementation.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -1,6 +1,17 @@
 module DeviseTokenAuth::Concerns::User
   extend ActiveSupport::Concern
 
+  def self.tokens_match?(token_hash, token)
+    @token_equality_cache ||= {}
+
+    key = "#{token_hash}/#{token}"
+    result = @token_equality_cache[key] ||= (BCrypt::Password.new(token_hash) == token)
+    if @token_equality_cache.size > 10000
+      @token_equality_cache = {}
+    end
+    result
+  end
+
   included do
     # Hack to check if devise is already enabled
     unless self.method_defined?(:devise_modules)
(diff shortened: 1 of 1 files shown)


nicolas-besnard:
What about the memory usage ?

booleanbetrayal:
should be negligible @nicolas-besnard ... the cache resets at 1000 entries and each entry uses < 100 bytes (83 fixed-length for the key) + boolean value. So ... less < 100k for the full collection.

booleanbetrayal:
FWIW - we saw about this reduce overhead from an additional ~75ms -> ~5ms (once cached) per request

nicolas-besnard:
Ok, you can do what ever you want, I don't care anymore about my memory :D

Nice PR ;)

booleanbetrayal:
Gonna run this a little longer in prod before merge, but so far, it's saving A LOT of time on our requests (we have token rotation disabled, FWIW).

booleanbetrayal:
Can't find any downside to this and it's speeding everything along beautifully. Merging!

lynndylanhurley:
Fantastic work, thanks @booleanbetrayal!!!

booleanbetrayal:
Thanks to @nbrustein  =]

irobayna:
:+1:

nicolas-besnard:
When are you bumping this gem to a new version ? I'd love to test this feature in production ;)

booleanbetrayal:
@nicolas-besnard - Just bumped and pushed 0.1.32.beta10 to RubyGems. Let me know if you have any issues!

Issue #121, opened 2015-01-26
User tokens don't properly deserialize


sdornan:
I am having trouble integrating devise_token_auth v0.1.31.beta9 into an existing Rails 4.2 app that already uses Devise.

I first ran the install. The user model was missing a few columns that devise_token_auth uses, so I edited down the generated migration to:

``` ruby
class DeviseTokenAuthCreateUsers < ActiveRecord::Migration
  def change
    add_column :users, :provider, :string, :null => false, :default => ""
    add_column :users, :uid, :string, :null => false, :default => ""
    add_column :users, :tokens, :text

    User.all.each do |user|
      user.provider = 'email'
      user.uid = user.email
      user.save!
    end

    add_index :users, [:uid, :provider],     :unique => true
  end
end
```

Attempting to perform the migration results in the following:

```
== 20150126183919 DeviseTokenAuthCreateUsers: migrating =======================
-- add_column(:users, :provider, :string, {:null=>false, :default=>""})
   -> 0.0218s
-- add_column(:users, :uid, :string, {:null=>false, :default=>""})
   -> 0.0164s
-- add_column(:users, :tokens, :text)
   -> 0.0012s
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `delete_if' for "{}":String/usr/local/Cellar/rbenv/0.4.0/versions/2.2.0/lib/ruby/gems/2.2.0/gems/devise_token_auth-0.1.31.beta9/app/models/devise_token_auth/concerns/user.rb:238:in `destroy_expired_tokens'
/Users/sam/Code/goals/db/migrate/20150126183919_devise_token_auth_create_users.rb:11:in `block in change'
/Users/sam/Code/goals/db/migrate/20150126183919_devise_token_auth_create_users.rb:7:in `change'
-e:1:in `<main>'
NoMethodError: undefined method `delete_if' for "{}":String
/usr/local/Cellar/rbenv/0.4.0/versions/2.2.0/lib/ruby/gems/2.2.0/gems/devise_token_auth-0.1.31.beta9/app/models/devise_token_auth/concerns/user.rb:238:in `destroy_expired_tokens'
/Users/sam/Code/goals/db/migrate/20150126183919_devise_token_auth_create_users.rb:11:in `block in change'
/Users/sam/Code/goals/db/migrate/20150126183919_devise_token_auth_create_users.rb:7:in `change'
-e:1:in `<main>'
Tasks: TOP => db:migrate
[...]

PR #276, opened 2015-06-24
Added json support for tokens

shicholas:
In the installation generator script, the migration template checks to see if the app is using postgres by seeing what sub-class of ``ActiveRecord::ConnectionAdapter`` the app is using. If it is postgres, the app will use a token column instead of a text one.

### DIFF
--- a/lib/generators/devise_token_auth/install_generator.rb
+++ b/lib/generators/devise_token_auth/install_generator.rb
@@ -115,5 +115,33 @@ module DeviseTokenAuth
       end
       match
     end
+
+    def json_supported_database?
+      (postgres? && postgres_correct_version?) || (mysql? && mysql_correct_version?)
+    end
+
+    def postgres?
+      database_name == 'ActiveRecord::ConnectionAdapters::PostgreSQLAdapter'
+    end
+
+    def postgres_correct_version?
+      database_version > '9.3'
+    end
+
+    def mysql?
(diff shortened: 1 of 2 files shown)


shicholas:
Since the tests run with SQLite, what do you think the best way to add tests are? I just did it manually.

shicholas:
hmm sorry it looks like my branch is out of sync. I'll update that soon

nicolas-besnard:
Could be nice to change column too ;)

shicholas:
fwiw, I got the inspiration from this comment: https://github.com/lynndylanhurley/devise_token_auth/issues/121#issuecomment-113790105

booleanbetrayal:
@shicholas - I think we need an additional check on `pg_types` as `json` was not supported until 9.3, and just checking the adapter is insufficient. see also: http://www.postgresql.org/docs/current/interactive/catalog-pg-type.html

Likewise, MySQL is now rolling `json` support, so we should account for that possibility as well. Not sure about SQLite at this point ...

shicholas:
okay sounds good, I will do a check for mysql and postgres with the right versions. I'll push up the some code soon. How should I test this though? Should we run the tests w/ those database versions?

shicholas:
pushed, let me know what you think.

booleanbetrayal:
@shicholas - LGTM ... would you mind rebasing and squashing your commits?

shicholas:
np, rebased and squashed.

booleanbetrayal:
thanks @shicholas !

PR #277, opened 2015-06-25
Update documentation for validate_token

### LINKED ISSUE #270: Shouldn't validate_token param be access-token, not auth_token?
I note that in the doc entry for validate_token the params specified are auth_token and uid. When trying to use the endpoint in cURL I can only make it work with uid, access-token, and client. Shouldn't the docs reflect this or am I missing something?

adamgall:
The `Usage TL;DR` section of README.md has documentation for `validate_token` that states that the method takes `uid` and `auth_token` as params. However, the class where `validate_token` is defined (`TokenValidationsController`) includes a `:before_filter :set_user_by_token`, who's definition parses headers and query params for `access-token`, not `auth_token`. This commit updates the documentation to specify that `access-token` is an accepted param instead of `auth_token`. This closes #270.


booleanbetrayal:
Thanks!

PR #296, opened 2015-07-08
Fix an issue for not :confirmable users

sebfie:
When we edit user password (after asking reset). Do not run skip_confirmation! is the user is not :confirmable

Also return json and 404 when user is not found


booleanbetrayal:
Thanks @sebfie !

PR #302, opened 2015-07-13
Set default provider to "email" in migration

### LINKED ISSUE #257: undefined method `provider' for #<User:0x007f49fd5da2e8>
I'm trying to setup the devise token auth for a rails application. I installed the gem and configured the roots and run those migrations. But I'm getting `undefined method `provider' for #<User:0x007f49fd5da2e8>` error while creating a user. 

```  
    NoMethodError in Devise::RegistrationsController#create
             undefined method `provider' for #<User:0x007f49fd5da2e8>

              Extracted source (around line #433):

          else
           match = match_attribute_method?(method.to_s)
           match ? attribute_missing(match, *args, &block) : super
          end
          end
```

and the route.rb is like this

```
   devise_for :admin_users, ActiveAdmin::Devise.config
   ActiveAdmin.routes(self)
   devise_for :users
   namespace :api do
     scope :v1 do
      mount_devise_token_auth_for 'User', at: 'auth'
     end
   end

```

schema

```
create_table "users", force: :cascade do |t|
t.string   "email",                  default: "", null: false
t.string   "encrypted_password",     default: "", null: false
t.string   "reset_password_token"
t.datetime "reset_password_sent_at"
t.datetime "remember_created_at"
t.integer  "sign_in_count",          default: 0,  null: false
t.datetime "current_sign_in_at"
t.datetime "last_sign_in_at"
t.string   "current_sign_in_ip"
t.string   "last_sign_in_ip"
t.datetime "created_at"
t.datetime "updated_at"
t.string   "authentication_token"
t.string   "provider",                            null: false
t.string   "uid",                    default: "", null: false
t.string   "confirmation_token"
t.datetime "confirmed_at"
t.datetime "confirmation_sent_at"
t.string   "unconfirmed_email"
t.string   "name"
t.string   "nickname"
t.string   "image"
t.string   "tokens"
end

   add_index "users", ["authentication_token"], name: "index_users_on_authentication_token", using: :btree
   add_index "users", ["email"], name: "index_users_on_email", unique: true, using: :btree
   add_index "users", ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true, using: :btree
   add_index "users", ["uid", "provider"], name: "index_users_on_uid_and_provider", unique: true, using: :btree

```

User model

```
      # Include default devise modules.
      devise :database_authenticatable, :registerable,
        :recoverable, :rememberable, :trackable, :validatable,
        :confirmable, :omniauthable
    has_many :movies
    has_many :reviews

   end

```

The POST request I used to signup.

```  
{
    "user": {
    "email": "email@g.com",
    "password": "password12345678",
    "password_confirmation": "password12345678" }

     }

```
and the response is like this.

```  
    {"status":"error","errors":["Please submit proper sign up data in request body."]}
 ```

What will be the issue behind that error, how can I solve it?

colavitam:
This change should fix incompatibility with standard devise controllers using default
migrations. Standard devise allows for provider to be null, but devise_token_auth requires it to have a value (as enforced by the default migration). The generated migration now sets this field to "email" by default. It should resolve #257 when the new migration is used. Setting the provider in the standard registrations controller is now redundant, but I left it as is to avoid breaking existing installations using the old migration.

### DIFF
--- a/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
+++ b/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
@@ -2,7 +2,7 @@ class DeviseTokenAuthCreate<%= user_class.pluralize %> < ActiveRecord::Migration
   def change
     create_table(:<%= user_class.pluralize.underscore %>) do |t|
       ## Required
-      t.string :provider, :null => false
+      t.string :provider, :null => false, :default => "email"
       t.string :uid, :null => false, :default => ""
 
       ## Database authenticatable


colavitam:
It may also be important to note that the scheme this app uses for setting the provider column isn't consistent with that used by standard devise. Changing the controllers and migrations to allow provider to be nil instead of "email" may make more sense, but it would likely make supporting users created by earlier versions more difficult.

booleanbetrayal:
This seems sane to me. Thanks for the PR @colavitam !

PR #307, opened 2015-07-16
Workaround: Add auth data to response body

richardfrost:
This is for devices that have trouble supporting CORS/Exposed Headers. On iOS we had an issue where it always removed any non-standard response headers, regardless of if they appeared in the exposed headers list. To work around this issue we add the auth data to the response body temporarily until ng-token-auth retrieves the necessary auth values and then removes it from the response body.

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -46,8 +46,11 @@ module DeviseTokenAuth
 
         yield if block_given?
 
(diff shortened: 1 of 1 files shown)


richardfrost:
This should be ready!

richardfrost:
Sorry about this - I failed to change the dropdown box when creating the pull request to keep it in our own repo. This was not intended to be a suggested fix, just a temporary workaround for our own uses. Thanks for the awesome project!

PR #308, opened 2015-07-17
Removing the unique emails validation

BrunoRomes:
__Removed the unique emails validation. It was conflicting with Devise's scoping login to subdomain.__

Explaining:
As seen in [Devise's wiki](https://github.com/plataformatec/devise/wiki/How-to:-Scope-login-to-subdomain), it is possible to scope the login to subdomains, allowing the creation of the resource with same email as many times as needed - as long as the subdomain changes. On the other hand, devise_token_auth had a unique email validation that kept this feature from working.
This validation was removed and anyone that wants to have unique emails must use Devise's **validatable** module.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -28,9 +28,6 @@ module DeviseTokenAuth::Concerns::User
     validates :email, presence: true, email: true, if: Proc.new { |u| u.provider == 'email' }
     validates_presence_of :uid, if: Proc.new { |u| u.provider != 'email' }
 
-    # only validate unique emails among email registration users
-    validate :unique_email_user, on: :create
-
     # can't set default on text fields in mysql, simulate here instead.
     after_save :set_empty_token_hash
     after_initialize :set_empty_token_hash


booleanbetrayal:
This sounds sane @BrunoRomes, but the PR as it stands is insufficient. We don't want to break existing, expected behavior. Please add `validatable` to the existing modules definition in `user.rb` and add a check for the module (`esource_class.devise_modules.include?(:validatable)`) when validating the email on create.

BrunoRomes:
In the `user.rb` the `validatable` module is already defined on line 19 (if the devise_modules is not defined), and I see no other place to define this module.
About checking this module `resource_class.devise_modules.include?(:validatable)` when creating the email, the `validatable` devise module already has this validation. Adding a validation like this in the code would just duplicate the same validation.
It may also be that I didn't fully understand your comment, if that's the case, please let me know.

booleanbetrayal:
Ah ... totally mis-grok'd the current module setup. Thanks for the explanation behind the existing validation. Could you just remove the `unique_email_user` method since it should no longer be in use?

BrunoRomes:
Sure.
One other thing I noticed is that there is two methods `email_changed?` and `email_required` that always returns false (so that Devise's email validation doesn't occur). Is this really the desired effect? Shouldn't Devise's validation take place?

booleanbetrayal:
It seems that if we're using `validatable` then we should be relying on it, and not short-circuiting any potential checks that are being done. That being said, we also have to play nice with OmniAuth flows, in addition to handling all the email-based auth use-cases. So those need to be regression tested as well.

Issue #285, opened 2015-07-02
Facebook omniauth redirection is missing url when testing on localhost


c0mrade:
Relating my previous issues :
https://github.com/lynndylanhurley/devise_token_auth/issues/166

The second part has been resolved regarding `api_user` and `user`. However first part still remains that I get redirected to a page without a hostname, just has a path.

Solution to this problem could be in the `OmniauthCallbacksController`, method `redirect_callbacks`, this line :
```
redirect_route = "/#{Devise.mappings[devise_mapping].as_json["path"]}/#{params[:provider]}/callback"
```

change it to :
```
redirect_route = "#{request.protocol}#{request.host_with_port}/#{Devise.mappings[devise_mapping].as_json["path"]}/#{params[:provider]}/callback"
```

Tried it locally works, any objections, different ideas how to do this?

c0mrade:
@lynndylanhurley can you see any issues with adding request protocol and port to the redirect_route?

booleanbetrayal:
@c0mrade - I don't see any issues, and it should probably account for the full URL info anyhow. Do you mind putting together a quick PR and making sure all existing tests pass?

c0mrade:
@booleanbetrayal sure, thanks for the response

booleanbetrayal:
Should be closed as of #322

PR #322, opened 2015-07-29
Fix invalid omniauth redirect

troggy:
Addresses issue #285


booleanbetrayal:
Thanks for this PR @troggy, but we shouldn't be using `File.join` since the separator isn't guaranteed to be the same across environments. Can you refactor to use vanilla string concatenation instead?

c0mrade:
:+1:

booleanbetrayal:
@c0mrade - between you and @troggy , whoever gets me a PR for this (without the `File.join` and preferably with a test) wins!  =]

troggy:
Like this?

c0mrade:
@booleanbetrayal is this good what troggy did?

booleanbetrayal:
Thanks! Got lost in the shuffle of a cross-country move.  =]

PR #323, opened 2015-07-31
feat(improved-omniauth): omniauth sameWindow and inAppBrowser flows

nbrustein:
This provides a mechanism to select between sameWindow (new default), newWindow (old default), or inAppBrowser modes for OAuth. This should allow for improved OAuth flows for a variety of use-cases, as well as supporting Cordova / PhoneGap implementations.

see also: lynndylanhurley/ng-token-auth#188 for ng-token-auth front-end implementation


chirag7jain:
would this work for native android apps ? if yes can you please add details within the readme..

nbrustein:
@chirag7jain I think that it would work in a native android app.  You would need to open the omniauth endpoint in an in-app browser and set the omniauth_window_type to 'inAppBrowser' (so something like `https:/myapp.com/auth/facebook?omniauth_window_type=inAppBrowser`).  That browser window will take the user to facebook and once they login it will forward them back to the omniauth callback endpoint of your rails app (something like `https://myapp.com/omniauth/facebook`).  

The page that is rendered by your rails app will have a requestCredentials function.  You will need to call that function from your native app to get the access token, client id, etc.

In order to call that requestCredentials function, listen for the loadstop event fired by the inAppBrowser (the loadstop event exists when using the inAppBrowser plugin in cordova/phone gap.  I assume it will exist for you as well).  When the loadstop event fires, use executeScript to call requestCredentials() in that in-app browser window.  The response will include all the info you need.  Then close the in-app browser window and use the token, etc. in subsequent requests.

It's not the easiest code to read, but you can see how ng-token-auth is doing this at the companion PR to this one: https://github.com/lynndylanhurley/ng-token-auth/pull/188/files  Look for the methods requestCredentialsViaExecuteScript and handleLoadStop.

Hopefully that makes some sense.  Let me know how it goes.

chirag7jain:
@ravi-ture please check this out

crysfel:
@nbrustein how can I integrate it with facebook ios sdk? I already have the facebook token on my app, but now I need to login on my server (which already implements this gem with ng-token-auth). Is there a tutorial or something?

Thanks

galharth:
Hey,
How can I use /auth/facebook with native iOS inAppBrowser (not Cordova/phonegap) ?
Thanks

geoandri:
Is there any news on this? Is it possible to login using iOS or android facebook SDKs and using the token to login to the rails server?

PR #329, opened 2015-08-08
Fixed error when using standard devise authentication

colavitam:
The auth_header and set_user_by_token code now keeps track of if the resource was previously set by devise. This avoids a crash when "change_headers_on_each_request" is set to false and standard devise is used for authentication. It also avoids a problem where the user can generate a new token with a custom client id if "change_headers_on_each_request" is true.


booleanbetrayal:
thanks for this PR @colavitam

PR #330, opened 2015-08-08
Tests to ensure standard devise has greater priority than tokens

colavitam:
This PR adds some tests to make sure devise_token_auth doesn't conflict with devise in some edge cases (namely when token information is provided alongside a standard devise session).


booleanbetrayal:
@colavitam - going to go ahead and take your word on this, as I don't have any experience with direct warden authentication outside of devise / devise_token_auth and several users are having issues ever since that support went in. Thanks for taking a dive on this!

Issue #340, opened 2015-08-17
Restricting access to controllers methods


gkopylov:
Hi, there.

The idea of this gem to improve security aspects of a token authentication is very nice, but some moments are bowl me out, for example there is almost no restriction access to controllers methods. It is a bad practice to make all methods public in controller, you can read about it here https://www.safaribooksonline.com/library/view/rails-cookbook/0596527314/ch04s13.html and here http://stackoverflow.com/questions/4495078/protected-and-private-methods-in-rails

Btw, I think there are some long methods that need to be refactored and we need to add rubocop to make the code cleaner.

PS. Is there any serious apps using this gem in production or it would be better to handle token authorization by yourself?(just curious).

lynndylanhurley:
@gkopylov - thanks for your feedback. We do accept pull requests.

gkopylov:
Ok, I have done some improvements but we still need to make this code better :-)

lynndylanhurley:
Thanks, we will review ASAP.

> Btw, I think there are some long methods that need to be refactored and we need to add rubocop to make the code cleaner.

I mostly agree with this, but let's proceed on a case-by-case basis. I worry sometimes about breaking up long functions just for the sake of it. I know that code climate complains about long functions, but there are also really good arguments for them. ([This article](http://number-none.com/blow/john_carmack_on_inlined_code.html) for example.)

PR #341, opened 2015-08-17
#340 Restrict access to controllers methods

gkopylov:
DeviseTokenAuth::OmniauthCallbacksController#omniauth_params is still need to be restricted but we need to create appropriate tests for this.

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -2,6 +2,7 @@ module DeviseTokenAuth
   class ApplicationController < DeviseController
     include DeviseTokenAuth::Concerns::SetUserByToken
 
+    protected
 
     def resource_class(m=nil)
       if m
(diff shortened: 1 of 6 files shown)


nbrustein:
@gkopylov is there a reason you did not also make omniauth_params a protected method?  (in omniauth_callbacks_controller)

gkopylov:
@nbrustein because it will raise an error for this test for example https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb#L75

So, becase of this I have written:
> but we need to create appropriate tests for this

nbrustein:
I see.  How would you feel about using `controller.send(:omniauth_params)` in the test.  Doesn't seem to crazy to me.

gkopylov:
It is a bit hackish but I think it is fair enough too.

Anyway I think it is better to handle token authorization by yourself(at least at the current moment)  :-)

booleanbetrayal:
This all looks good to me!  :+1:

PR #347, opened 2015-08-19
Fully support OmniauthCallbacksController action overrides. Fixes #186.

### LINKED ISSUE #186: Unable to override OmniauthCallbacksController#redirect_callbacks
I noticed that my `OmniauthCallbacksController#redirect_callbacks` override implementation wasn't being used. I dug into this a bit, and it turns out that the override for `omniauth_callbacks` will never touch this particular route, as it is defined in [config/routes.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/master/config/routes.rb#L3) (and not [lib/devise_token_auth/rails/routes.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/master/lib/devise_token_auth/rails/routes.rb#L51)).

I was able to fix this by moving the route definition to the latter file (where the overrides are considered).

    match "#{DeviseTokenAuth.omniauth_prefix}/:provider/callback", controller: omniauth_ctrl, action: "redirect_callbacks", via: [:get]

Does this present other issues? Or can we safely move this route definition and remove `config/routes.rb`?

I can put a PR together with the update if we're okay with moving forward.

tbloncar:
Remove `config/routes.rb` and define all routes in `lib/devise_token_auth/rails/routes.rb`; this ensures that all `OmniauthCallbacksController` actions can be overridden.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,0 @@
-Rails.application.routes.draw do
-  if defined?(::OmniAuth)
-    get "#{DeviseTokenAuth.omniauth_prefix}/:provider/callback", to: "devise_token_auth/omniauth_callbacks#redirect_callbacks"
-    get "#{DeviseTokenAuth.omniauth_prefix}/failure", to: "devise_token_auth/omniauth_callbacks#omniauth_failure"
-  end
-end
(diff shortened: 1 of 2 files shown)


booleanbetrayal:
Thanks @tbloncar !

tbloncar:
:+1:

PR #349, opened 2015-08-21
Add documentation to explain gotcha with rails-api.

aarongray:
If you are using rails-api instead of vanilla rails, the concern in the application controller does not get created automatically. Since rails-api has been merged into rails and this is a sneaky little gotcha, I propose we add a heads-up to the readme to reduce head-banging for newcomers to this gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ The following events will take place when using the install generator:
 
 * Routes will be appended to file at `config/routes.rb`. [Read more](#mounting-routes).
 
-* A concern will be included by your application controller at `app/controllers/application_controller.rb`. [Read more](#controller-methods).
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
Thanks @aarongray . Merged, but maybe we'd like to support rails-api in the installation script. Not too familiar ... Is there a base class like ApplicationController that would be a predictable hook?

aarongray:
That would be even better!

Rails-api apps have an ApplicationController just like vanilla rails apps, except in rails-api, it inherits from ActionController::API, instead of the usual ActionController::Base.

aarongray:
Okay, how about [this](https://github.com/lynndylanhurley/devise_token_auth/pull/350)?

booleanbetrayal:
:+1:

PR #350, opened 2015-08-21
Fix concern not being inserted for rails-api apps.

aarongray:
Rails-api apps have an ApplicationController just like vanilla rails apps, except in rails-api, it inherits from ActionController::API, instead of the usual ActionController::Base.


booleanbetrayal:
Looks great. Thanks @aarongray !

aarongray:
Glad to help. Big thanks to you guys for the handy library. ^_^

PR #352, opened 2015-08-21
Allowing devise patch version updates

### LINKED ISSUE #351: Allow devise patch version updates
The most current 'devise' version is `3.5.2` but this gem specifies the exact version `3.5.1` which is too specific in my opinion. I can understand that there is the will to test things before upgrading but as this is a security aware and enhancing gem I feel it really should allow security patches and bug fixes to devise be immediately available.

I suggest using the `"~> 3.5.1"` specification style as this would allow others use 3.5.2 but not 3.6 until it is tested to work with 3.6+ versions.

Dschee:
I have just changed the line specifying the exact version 3.5.1 of devise to be `"~> 3.5.1"` to allow for devise patch version updates. This fixes #351.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -34,7 +34,7 @@ PATH
   remote: .
   specs:
     devise_token_auth (0.1.34)
-      devise (= 3.5.1)
+      devise (~> 3.5.1)
       rails (~> 4.2)
 
 GEM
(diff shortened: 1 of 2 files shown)


Dschee:
The changes from this pull request were inserted by @booleanbetrayal in bf2cc1b69a1456e53b40de87ec98245102d235a8.

PR #356, opened 2015-08-25
Add description to readme about the devise.rb initializer.

aarongray:
Right now the readme doesn't contain any info about this file, although it is an important part of configuring this library, and the [example app uses it](https://github.com/lynndylanhurley/devise_token_auth_demo/blob/master/config/initializers/devise.rb). Adding this section here should make it a little easier on newcomers to learn about this part of configuring devise.


booleanbetrayal:
thanks @aarongray !

aarongray:
:+1:

PR #370, opened 2015-09-01
Allow authentication with an API Token 

neutronz:
Cover the case where an api consumer (some 3rd party organization) has been given an api key/token rather than email/password that will allow access to the api.

The default is that the request header "Authentication" is used for the token, however, its configurable in the setup block.


akshah123:
:+1:

bobmshannon:
Are there any plans in the future to merge this in and make it a supported feature?

PR #371, opened 2015-09-02
fix(oauth): fixes #368: do not serialize the entire user object in the url when redirecting from oauth

### LINKED ISSUE #368: omniauth: when redirecting, user object should not be serialized into url
in `render_data_or_redirect`, we only want to include the user object in the data when using `newWindow` or `inAppBrowser`.  In the `sameWindow` case, ng-token-auth is going to make separate api call anyway to load up the user.

This defends against problems caused by a big user object leading to a big url.

It could be a breaking change for people not using ng-token-auth

nbrustein:
fixes #368 

ng-token-auth is not using this data, so this should be fine for anyone using ng-token-auth.  It is potentially a breaking change if anyone not using ng-token-auth was relying on it, but this functionality has only been around for a few weeks, so hopefully nobody is.

### DIFF
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -39,7 +39,7 @@ module DeviseTokenAuth
 
       yield if block_given?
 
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
LGTM!

PR #376, opened 2015-09-14
Semantic Registrations Error Status Codes

occamin:
I noticed that the registration controller has some unexpected status codes, namely `403 forbidden`, where `400 bad_request` or `422 unprocessable_entity` were more appropriate.

I also found a few opportunities to simplify or clarify the code and updated the tests

### DIFF
--- a/app/controllers/devise_token_auth/registrations_controller.rb
+++ b/app/controllers/devise_token_auth/registrations_controller.rb
@@ -10,17 +10,10 @@ module DeviseTokenAuth
       @resource.provider   = "email"
 
       # honor devise configuration for case_insensitive_keys
-      if resource_class.case_insensitive_keys.include?(:email)
-        @resource.email = sign_up_params[:email].try :downcase
-      else
-        @resource.email = sign_up_params[:email]
-      end
+      @resource.email.try :downcase! if resource_class.case_insensitive_keys.include?(:email)
 
-      # give redirect value from params priority
-      @redirect_url = params[:confirm_success_url]
-
-      # fall back to default value if provided
-      @redirect_url ||= DeviseTokenAuth.default_confirm_success_url
+      # give redirect value from params priority or fall back to default value if provided
+      @redirect_url = params[:confirm_success_url] || DeviseTokenAuth.default_confirm_success_url
 
       # success redirect url is required
       if resource_class.devise_modules.include?(:confirmable) && !@redirect_url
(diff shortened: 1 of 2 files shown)


app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -10,17 +10,10 @@ def create
       @resource.provider   = "email"
 
       # honor devise configuration for case_insensitive_keys
-      if resource_class.case_insensitive_keys.include?(:email)
-        @resource.email = sign_up_params[:email].try :downcase
-      else
-        @resource.email = sign_up_params[:email]
-      end
```
occamin: The `email` attribute will already be set if it was passed in as a parameter. We can just try and downcase it when required

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -10,17 +10,10 @@ def create
...
+      @resource.email.try :downcase! if resource_class.case_insensitive_keys.include?(:email)
```
occamin: The `email` attribute will already be set if it was passed in as a parameter. We can just try and downcase it when required.

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -10,17 +10,10 @@ def create
...
 
-      # give redirect value from params priority
-      redirect_url = params[:confirm_success_url]
-
-      # fall back to default value if provided
-      redirect_url ||= DeviseTokenAuth.default_confirm_success_url
+      # give redirect value from params priority or fall back to default value if provided
+      redirect_url = params[:confirm_success_url] || DeviseTokenAuth.default_confirm_success_url
```
occamin: This is a bit more concise without loss of legibility

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -28,68 +21,64 @@ def create
           status: 'error',
           data:   @resource.as_json,
           errors: [I18n.t("devise_token_auth.registrations.missing_confirm_success_url")]
-        }, status: 403
+        }, status: :bad_request
```
occamin: `bad_request` seems the most appropriate given the missing parameter. There is a good discussion on [stackoverflow](http://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data) about `400` vs `422`

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -28,68 +21,64 @@ def create
...
-            errors: [I18n.t("devise_token_auth.registrations.redirect_url_not_allowed", redirect_url: redirect_url)]
-          }, status: 403
-        end
+      if DeviseTokenAuth.redirect_whitelist &&
+         !DeviseTokenAuth.redirect_whitelist.include?(redirect_url)
+        return render json: {
+          status: 'error',
+          data:   @resource.as_json,
+          errors: [I18n.t("devise_token_auth.registrations.redirect_url_not_allowed", redirect_url: redirect_url)]
+        }, status: :forbidden
```
occamin: This is logically equivalent and manages to undo one level of indentation

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -28,68 +21,64 @@ def create
...
       end
```
occamin: Same as above, based on the [ruby styleguide](https://github.com/bbatsov/ruby-style-guide#begin-implicit)

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -28,68 +21,64 @@ def create
...
-            @resource.send_confirmation_instructions({
-              client_config: params[:config_name],
-              redirect_url: redirect_url
-            })
-
-          else
-            # email auth has been bypassed, authenticate user
-            @client_id = SecureRandom.urlsafe_base64(nil, false)
-            @token     = SecureRandom.urlsafe_base64(nil, false)
+      # override email confirmation, must be sent manually from ctrl
```
occamin: Same as the above, based on the [ruby styleguide](https://github.com/bbatsov/ruby-style-guide#begin-implicit)

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -142,11 +131,8 @@ def account_update_params
     private
 
     def resource_update_method
-      if DeviseTokenAuth.check_current_password_before_update == :attributes
-        "update_with_password"
-      elsif DeviseTokenAuth.check_current_password_before_update == :password and account_update_params.has_key?(:password)
-        "update_with_password"
-      elsif account_update_params.has_key?(:current_password)
+      if DeviseTokenAuth.check_current_password_before_update == :attributes ||
+         account_update_params.has_key?(:current_password)
```
occamin: This was logically redundant
```
elsif DeviseTokenAuth.check_current_password_before_update == :password and account_update_params.has_key?(:password)
  "update_with_password
```
since it's automatically covered by
```
elsif account_update_params.has_key?(:current_password)
  account_update_params.has_key?(:current_password)
```

Combined the clauses into one for brevity.

lynndylanhurley:
Thanks @occamin - I'll review this ASAP!!

occamin:
Hey @lynndylanhurley, 

No problem. If you happy with the general changes I'm happy to get the latest and fix any merge conflicts.

booleanbetrayal:
@occamin - could you go ahead and rebase against master and resolve those conflicts?

occamin:
@booleanbetrayal - done

booleanbetrayal:
@occamin  - I definitely like all the code cleaning tweaks, and the status code changes seem reasonable to me ... the only issue I'm hung up on is the breaking nature of the change to any client libraries. I'm almost thinking that those status codes need to be configurable in order to support clients in the wild (several hybrid apps currently use this gem for auth, etc). Do you have any strong opinions @lynndylanhurley ?

occamin:
@booleanbetrayal - that's a fair point, and I think there is value is having customisable responses both in terms of backwards compatibility and consistency with various json response formats the user's api might be using (e.g. jsonapi.org).

From a breaking change point of view I suspect the impact would be less than expected because we already distinguish 403s from 404s, but 403's include a variety of other errors unrelated to access (e.g. bad request, or unprocessable data) so I suspect clients or either handling non 2xx responses generically or distinguishing between errors based on error message instead of status code.

I noticed since my first pull request, responses have been been out into separate methods which would have likely solved my initial problem since I could just inherit and easily override them in my own controller. The only real changes needed are for some instance variables (@resource and @redirect_url) to be protected or passed in as parameters so that subclasses have access to them

zachfeldman:
@occamin thanks for your contribution, but as you mentioned I think people can override where they'd like to at this point and there's many merge conflicts. I am going to close this up for now.

occamin:
@zachfeldman Sure thing, didn't realise this was even still open.

Issue #127, opened 2015-02-01
Would it be useful to integrate login with username ?


nicolas-besnard:
Has mentionned in #126

lynndylanhurley:
I just scanned the devise documentation and found this:
https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address

So it is possible with standard devise, but the process seems to be rather involved. Also, I'm not sure if this gem will require any modifications to accommodate for this kind of setup.

And in addition to following the steps from the wiki, a new config option will need to be added to ng-token-auth to specify which params are sent to the sessions controller (i.e. `username` instead of `email`).

I think the best course of action is to add a test suite that follows the example in the devise wiki. If it works, then that's great, and if not then we can modify this gem as is needed. I'm short on time right now, but I'll get around to it as soon as I can.

nicolas-besnard:
I'm available to do it, but I don't use to work with `MiniTest` :( I'll try tomorrow to work with ``MiniTest`

lynndylanhurley:
Thanks @nicolas-besnard - I appreciate any help I can get :smiley:

To get the dev environment set up, you can follow [these steps](https://github.com/lynndylanhurley/devise_token_auth#contributing).

I recommend the following approach to testing this feature:

1. Create a new model and migration in the [test app](https://github.com/lynndylanhurley/devise_token_auth/tree/master/test/dummy) that follows the [example](https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address) in the devise wiki. The existing models are [here](https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address) - you can just follow those for examples.
2. Create tests for each of the cases described in the wiki example. This includes:
  * registration ([link to existing tests](https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/controllers/devise_token_auth/registrations_controller_test.rb))
  * confirmation ([link](https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/controllers/devise_token_auth/confirmations_controller_test.rb))
  * sign in ([link](https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/controllers/devise_token_auth/sessions_controller_test.rb))
[...]

PR #395, opened 2015-10-04
Implement login auth key according to Devise wiki

luciomartinez:
Devise is [capable](https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address) to use email, username or both fields for auth. Currently, Devise Token Auth supports email or username but not both at the same time. This PR is an intention to start the implementation of `login` as an auth key to use the previous fields simultaneously.

This is rather an ad-hoc implementation that makes possible to send a payload with `login` attribute with a valid username or email as its value.

**Important:** this is a proposal based on #127 from which I was inspired for this implementation.

In the few tests written, it seems to be working as expected, but I'm pretty sure that the _SessionsController_ modifications are mostly hard-coded due my lack of knowledge for the field `nickname` on the project.

I'll try to make all the necessary improvements with a bit of guidance.
Please, feel free to throw any critic here and I hope this makes any sense :)

Thanks.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -5,9 +5,17 @@ module DeviseTokenAuth::Concerns::SetUserByToken
   included do
     before_action :set_request_start
     after_action :update_auth_header
+
+    before_action :default_permitted_parameters, if: :devise_controller?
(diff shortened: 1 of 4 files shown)


zachfeldman:
Hi there @luciomartinez    ,

Hope all is well.

PR #404, opened 2015-10-12
Drop .ruby-version file

xymbol:
If needed, add a sample file or `required_ruby_version` entry.


booleanbetrayal:
I don't have strong opinions about this, but seeing as how most gemset managers can parse Gemfile / gemspecs I don't see any harm. One less thing to maintain.  =]

xymbol:
@booleanbetrayal Agree but, it's a common practice to don't have it around.

lynndylanhurley:
@xymbol - If we're going to merge this, can you please add the file to the `.gitignore` list as well?

xymbol:
@lynndylanhurley Sorry about that, done.

PR #407, opened 2015-10-14
Increment number of sign in and date of sign in after confirmation. S…

rmvenancio:
I made 3 changes: 
. Added pt-PT to rep 
. When the user follow the confirmation link, he is automatically logged in. But the login date and login number increment is not updated.  Added sign_in(@resource) to update this information 
. When you visit the change password link for the second time (sent by email), you should be redirected to a 404 page instead of get a window with a json message.

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -15,8 +15,10 @@ module DeviseTokenAuth
           expiry: expiry
         }
 
-        @resource.save!
+        sign_in(@resource)
 
+        @resource.save!
+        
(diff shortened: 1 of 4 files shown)


booleanbetrayal:
Thanks @rmvenancio! Looks like there are some failing tests that would need to be fixed, though. Would you also be able to split this up into 3 separate PRs, as well?

rmvenancio:
Hi @booleanbetrayal. I'm new to rails and this is my first collaboration. I will try to to find how to make the tests to pass and split this in 3 different PR. BR.

rmvenancio:
Will create different PRs

PR #410, opened 2015-10-16
Flag signin when user confirms email address.

rmvenancio:
When the user follows the confirmation link with success, he automatically sign in to the system.

The information about the number of sign in, the ip etc is not recorded in the database.
This change allows this information to be saved to the database.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -150,6 +150,7 @@ GEM
     multi_json (1.11.2)
     multi_xml (0.5.5)
     multipart-post (2.0.0)
+    mysql2 (0.3.19)
     nenv (0.2.0)
     nokogiri (1.6.6.2)
       mini_portile (~> 0.6.0)
@@ -250,6 +251,7 @@ DEPENDENCIES
   minitest-rails
   minitest-reporters
   mocha
+  mysql2
   omniauth-facebook!
   omniauth-github!
   omniauth-google-oauth2!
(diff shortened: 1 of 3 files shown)


booleanbetrayal:
@rmvenancio - Sorry for the delay on this. Barely managing to keep above water. Agree that this should happen as described. Could you add some specs around the sign in increment etc?

rmvenancio:
Hi @booleanbetrayal, 

No problem. I'm also with several deliverables until the end of the year. I will create some specs but I can't promise nothing until the end of the year.

booleanbetrayal:
@rmvenancio - We be using `sign_in` consistently across controllers (ie - SessionsController) if we decide to make this change.

rmvenancio:
Hi @booleanbetrayal, 
Not sure if I understood your message. 

If you don't agree with that solution you can reject this PR. No problem at all. 

I think the confirmation controller have two problems: 
1st: When a user follows the confirmation link, he is automatically logged in the system but some information is not recorded (sign_in_count, current_sign_in_at, last_sign_in_at, current_sign_in_ip and last_sign_in_ip). The call of sign_in in the confirmation controller was the (fast) way I found to solve this problem. 

2nd: The user can follow the confirmation link unlimited times and log in into the system without using a password. Possible solution was #519 in order for this solution to work, we will need to solve the 1st issue. 

Is up to you the acceptance of this PR. What you decide for me is ok. For my project I need this kind of behaviours.

zachfeldman:
@rmvenancio nice PR.

@lynndylanhurley this doesn't seem too controversial, so I've marked it easy-to-merge.

PR #411, opened 2015-10-16
404 for invalid link with password reset token

rmvenancio:
When you visit the change password link for the second time (sent by email), you should be redirected to a 404 page instead of get a window with a json message.

This behaviour is also required when you have a wrong password confirmtion link.


lynndylanhurley:
Looks good, thanks @rmvenancio!!

rmvenancio:
You're welcome @lynndylanhurley. Best regards.

PR #417, opened 2015-10-24
Change default message for already in use error and added to english …

ponyesteves:
…and spanish translation files (en.yml and es.yml)


lynndylanhurley:
Thanks @ponyesteves!!!

ponyesteves:
You're welcome and thanks @lynndylanhurley , for the great job on this gem and it's angular library.  I'm implementing both  on a rails angular spanish proyect, so i guess i'll be contributing with some other translations as well!

PR #421, opened 2015-10-27
fix(url): preserve query parameters when building urls

nbrustein:
When logging in with omniauth, we need to auth_origin_url to include the query params from the original page.  Seems like most people would want this behavior, although in some circumstances this could conceivably be a breaking change.


booleanbetrayal:
This looks good to me. Will add a note in the changelog when cutting a new gem.

PR #428, opened 2015-10-30
feat(enable-standard-devise): allow configurable support of legacy Devise authentication

### LINKED ISSUE #297: undefined method `tokens' for #<Hash:0x000000063f0920>
This part is triggering this exception, because `warden.user(rc.to_s.underscore.to_sym)` returns hash, not User object. What should I do?
https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L32

booleanbetrayal:
Fixes #297

BREAKING CHANGE: Updates legacy Devise support to default to disabled rather than enabled.

This was causing all sorts of random issues for people who may not have needed the integration. This isolates it and denotes it as an experimental feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,18 @@
 
 ## Features
 
+- **Standard Devise**: Allow conditional support of legacy Devise. Now defaults to disabled.
+- **Localization**: Add German translation(de)
 - **Batch Requests**: Prevent batching of requests by appending `unbatch=true` param to request URL
+
+## Fixes
+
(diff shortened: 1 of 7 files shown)


lynndylanhurley:
This looks great @booleanbetrayal!

booleanbetrayal:
Yay!

booleanbetrayal:
left out a revert on the `bypass` flag that was clearing the `before_filter` stuff -- see also: e0b4dff

PR #430, opened 2015-11-02
Raise error in controller method

ArneZsng:
Upgrading from v0.1.36 to v0.1.37.beta3 I noticed a change in passwords_controller.rb that broke at least my code with regards to #411

Though I agree that it should not show a json object in the browser, our logic and I guess some other projects as well might have overwritten render_edit_error to redirect to a single page app with a more specific error message.

I readded the render_edit_error method to easily override the behavior.


booleanbetrayal:
Seems reasonable. Thanks @ArneZsng

PR #431, opened 2015-11-02
Limited number of devices

paulosoares86:
@booleanbetrayal, I'm happy to show you my PR. My intention is to allow users to configure the max number of allowed devices. In my API, for example, I don't want concurrent devices, so I'll set max_number_of_clients = 1. In tipical applications, however, I think 10 is a reasonable value. 

Using this configuration we can also avoid the max length problem in database and eventually a DoS attack.

Please take a look and tell me what you think.


booleanbetrayal:
@paulosoares86 - haven't gotten to play with it, but the code looks good. could you add some tests that show the tokens queue gets cycled as expected? A rebase with squashed commits would be ideal. Thanks!

paulosoares86:
No problem, @booleanbetrayal. I've changed the parameter name, included some tests and a short description of parameter in readme.

You are welcome!

booleanbetrayal:
Hey @paulosoares86 - I'm seeing all sorts of merge commits and duplicated commits. If you're unable to squash those in a rebase, maybe just creating a new branch for the PR and applying your patch? Would like for you to get the credit if this gets into master, but would like to keep the commit history clean.

paulosoares86:
I'm sorry, @ booleanbetrayal. I've made a rebase, but it seems better to make another branch with only one commit.

booleanbetrayal:
@paulosoares86 - Whatever's easiest! Just let me know if you open up a new PR.

PR #435, opened 2015-11-07
Device mapping duplicate method removed

chirag7jain:
Since SetUserByToken concern already has the resource_class method and
application controller is including the same concern, there is no need to
keep the same duplicate method in application controller.

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -1,17 +1,5 @@
 module DeviseTokenAuth
   class ApplicationController < DeviseController
     include DeviseTokenAuth::Concerns::SetUserByToken
-
-    protected
-
-    def resource_class(m=nil)
-      if m
-        mapping = Devise.mappings[m]
-      else
-        mapping = Devise.mappings[resource_name] || Devise.mappings.values.first
-      end
-
-      mapping.to
-    end
   end
 end


chirag7jain:
pull request https://github.com/lynndylanhurley/devise_token_auth/pull/435

booleanbetrayal:
@chirag7jain - can you please fix the specs in an amended commit?

chirag7jain:
@booleanbetrayal bit unsure what you mean by specs...

booleanbetrayal:
@chirag7jain - the project's unit tests which can be run via `guard`

lolgear:
@booleanbetrayal well, it is not a @chirag7jain fault.

 I try the same and see interesting results:

Nearly all tests are fail with  `ArgumentError:         ArgumentError: wrong number of arguments (1 for 0)`
```
ERROR["test_response_should_fail", #<Class:0x007fe3dc820768>, 2015-11-19 00:37:35 +0300]
 test_response_should_fail#DeviseTokenAuth::PasswordsController::Password reset::change password::password mismatch error (1447882655.67s)
ArgumentError:         ArgumentError: wrong number of arguments (1 for 0)
```

lolgear:
@booleanbetrayal ah, I see, it can't be ripped from ApplicationController. The father of ApplicationController (DeviseController) has method

```
  def resource_class
    devise_mapping.to
  end
```

booleanbetrayal:
going to close this PR, but feel free to re-open another if you think there is a similar issue

PR #436, opened 2015-11-07
to keep coherent with devise. pt instead of pt-PT.yml

rmvenancio:


booleanbetrayal:
thanks for this!

rmvenancio:
You're welcome!

PR #443, opened 2015-11-16
Error message

rmvenancio:
Added I18n for an error message when an email is already registered


app/controllers/devise_token_auth/passwords_controller.rb
```diff
@@ -100,7 +100,7 @@ def edit
           config:         params[:config]
         }))
       else
-        render_edit_error
+        raise ActionController::RoutingError.new('Not Found')
```
rmvenancio: Please ignore this line

rmvenancio:
Please take only in consideration the files: 
app/models/devise_token_auth/concerns/user.rb
config/locales/en.yml
config/locales/pt-BR.yml
config/locales/pt.yml

I've used my rep that have some differences from this one. Thanks

Issue #452, opened 2015-11-24
cancancan: load_and_authorize_resources causes method_missing failure


lolgear:
setup:

```
cancancan-1.13.1
rails-4.2.5
devise_token_auth - commit #df960d9409dcd4476a3f6694e8bf30e57000599a

```

```
class DevicesController < ApiBaseController
    before_filter :output_user
  # load_and_authorize_resource
  inherit_resources
  respond_to :json

  def output_user
    puts "current user is: #{@current_user}"
  end

  def index     
    @devices = Device.all.sorted_by(params[:sort])
    render json: @devices, status: :ok
  end
end
```

works fine, but when I try to uncomment `load_and_authorize_resources`, I got error:

```
NoMethodError (undefined method `find_by_uid' for #<Class:0x007f8499a78670>):
  activerecord (4.2.5) lib/active_record/dynamic_matchers.rb:26:in `method_missing'
  devise_token_auth/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:71:in `set_user_by_token'
  devise_token_auth/lib/devise_token_auth/controllers/helpers.rb:116:in `current_user'
  cancancan (1.13.1) lib/cancan/controller_additions.rb:357:in `current_ability'
  cancancan (1.13.1) lib/cancan/controller_resource.rb:214:in `current_ability'
  cancancan (1.13.1) lib/cancan/controller_resource.rb:72:in `load_collection?'
  cancancan (1.13.1) lib/cancan/controller_resource.rb:33:in `load_resource'
  cancancan (1.13.1) lib/cancan/controller_resource.rb:25:in `load_and_authorize_resource'
  cancancan (1.13.1) lib/cancan/controller_resource.rb:10:in `block in add_before_filter'
```

I try to fix resource_class collision bug by:

```
    def set_user_by_token(mapping=nil)
    # ......
    rc = resource_class(mapping) rescue resource_class

    # .......

    puts "this class is: #{rc}"
    user = uid && rc.find_by_uid(uid)
```

But when I try to access to /devices.json, I see:

```
this class is: Device
Completed 500 Internal Server Error in 3ms (ActiveRecord: 0.7ms)

NoMethodError (undefined method `find_by_uid' for #<Class:0x007f849b67d578>):
```

lolgear:
#454

zachfeldman:
Hi there @lolgear    ,

[...]

Issue #450, opened 2015-11-23
DeviseTokenAuth::Concerns::SetUserByToken wrong number of arguments 1 for 0


lolgear:
Something strange happens inside DeviseTokenAuth::Concerns::SetUserByToken#set_user_by_token()

```
  activesupport (4.2.5) lib/active_support/core_ext/class/attribute.rb:106:in `block (2 levels) in class_attribute'
devise_token_auth/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:24:in `set_user_by_token'
  devise_token_auth/lib/devise_token_auth/controllers/helpers.rb:115:in `current_user'
  devise_token_auth/lib/devise_token_auth/controllers/helpers.rb:103:in `authenticate_user!'
```

It seems that `rc = resource_class(mapping)` doesn't work well. ( `resource_class` doesn't work well with parameter )

lolgear:
I think that they are related: https://github.com/lynndylanhurley/devise_token_auth/issues/433

lolgear:
additional info:

```
activesupport (4.2.5) lib/active_support/core_ext/class/attribute.rb:106:in `block (2 levels) in class_attribute'
devise_token_auth (0.1.36) app/controllers/devise_token_auth/concerns/set_user_by_token.rb:21:in `set_user_by_token'
devise_token_auth (0.1.36) lib/devise_token_auth/controllers/helpers.rb:115:in `current_user'
cancancan (1.13.1) lib/cancan/controller_additions.rb:357:in `current_ability'
cancancan (1.13.1) lib/cancan/controller_resource.rb:214:in `current_ability'
cancancan (1.13.1) lib/cancan/controller_resource.rb:72:in `load_collection?'
cancancan (1.13.1) lib/cancan/controller_resource.rb:33:in `load_resource'
cancancan (1.13.1) lib/cancan/controller_resource.rb:25:in `load_and_authorize_resource'
```

reaper:
Got the same issue on master (9668dc77a0e9) when authenticating a user not omniauthable

config.rb
```ruby
devise_for :admins, controllers: { sessions: 'admin/sessions' }
mount_devise_token_auth_for 'User', at: 'users_auth'
```

GET: http://localhost:3000/admins/sign_in
produces:

```
ArgumentError - wrong number of arguments (1 for 0):
[...]

PR #454, opened 2015-11-26
concerns: set_user_by_token method collision possible fix

lolgear:
Anonymous function added as rescue to resource_class method collision. ( fix 'missing method' errors with cancan usage )

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -18,10 +18,18 @@ module DeviseTokenAuth::Concerns::SetUserByToken
   # user auth
   def set_user_by_token(mapping=nil)
     # determine target authentication class
-    rc = resource_class(mapping)
+    rc_class = ->(m=nil) do
+      if m
+        Devise.mappings[m]
+      else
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
hey @lolgear - could you add a comment that also references #452 in that workaround block? just think this will be one of those situations where we come back awhile later and have no idea why this was done.  =]

lolgear:
@booleanbetrayal, I want to write specs for tihs. It seems that `resource_class` doesn't work well for some situations.
The first try was: remove resource_class duplicate (failed): https://github.com/lynndylanhurley/devise_token_auth/pull/435

I commented why errors occurs.

booleanbetrayal:
Specs would be great. Let me know once they're in and I'll take a look at this PR.

booleanbetrayal:
Hey @lolgear - you want to see this one through?

lolgear:
Hi, @booleanbetrayal - yes, I want

lolgear:
@booleanbetrayal but it seems, that @rossshannon improve my PR and find another ( maybe better ) solution. ( look at issue #450 that was referenced earlier )

jvalenciag:
+1

PR #466, opened 2015-12-12
Fix locales `errors.messages.already_in_use` + clean up

### LINKED ISSUE #463: locales `errors.messages.already_in_use` seems broken
I am using devise_token_auth v.0.1.36 in a RESTful API only server application.

My default i18n locale is `:fr`.

While testing my application with RSpec, I keep having devise token auth error messages in english while the rest (devise and rails error messages) are correctly displayed in french.

It happens on the `already_in_use` key, which is defined in `errors.messages.already_in_use` in the locale file and called [here](https://github.com/lynndylanhurley/devise_token_auth/blob/543c477fe5cd7f8d1de16ef2c7a1d0611c461772/app/models/devise_token_auth/concerns/user.rb#L242).

By the way, I noticed that devise_token_auth doesn't _really_ follow (which seems to be a convention among rails gems) the schema : `errors.messages.<your_key>` inside the locale file. I think it would be more consistent to follow this rule. What do you think ?

ValentinTrinque:
fixes #463


ValentinTrinque:
I have tested it, calling my own repo and branch from my Gemfile. It works great.

The locale is correctly changed.

booleanbetrayal:
This looks great @ValentinTrinque . Thanks!

PR #483, opened 2015-12-22
Locale file for both zh-TW and zh-HK

TravisTam:
There may be some wording not fully equal taiwan usage.


booleanbetrayal:
Great! Thanks @TravisTam !

PR #490, opened 2016-01-05
Update version.rb

charlesdg:
Can you update your gem version with the changes you accept some days ago?


booleanbetrayal:
updated and pushed as `0.1.37`

Issue #397, opened 2015-10-05
Errors after removing confirmable


Andrew-Max:
seems exactly like : https://github.com/lynndylanhurley/devise_token_auth/issues/58 although that issue was marked as resolved and I still see the error.

I removed `devise :confirmable`

and removed the confirmable lines from the migration by hand. Afterwards i see `NameError: undefined local variable or method `confirmed_at'`

as suggested in the other issue, adding the lines

```
  def confirmed_at
      DateTime.now
  end

  def confirmed_at=(val)
  end
```

To the my user model solves the problem but I would prefer to avoid this give that it is hacky

fabiancarlos:
Yeah, mee to! Someone solved?

MarketingAjay:
@Andrew-Max @fabiancarlos I just ran into the same issue.  The problem happens if your `include` statement is BEFORE the `devise` statement.  This is unfortunate because the generator puts it at the top of the model by default.  Switch the order and you should be good.

```
class User < ActiveRecord::Base
  devise :database_authenticatable, :registerable,
         :recoverable, :rememberable, :trackable, :validatable

  include DeviseTokenAuth::Concerns::User
end
```

jeryRazakarison:
Hi, I have a different error, while **removing :confirmable** from the `devise` statement.
I'm on Rails 5 beta, using Devise master branch, and devise_token_auth master branch.

When I make the post request, I have an error 500

    Unpermitted parameter: registration
    Completed 500 Internal Server Error in 78ms (ActiveRecord: 0.0ms)

And here is the stack
```
ArgumentError (After create callback :send_on_create_confirmation_instructions has not been defined):
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:641:in `block (2 levels) in skip_callback'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:637:in `each'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:637:in `block in skip_callback'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:568:in `block in __update_callbacks'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:566:in `reverse_each'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:566:in `__update_callbacks'
  activesupport (5.0.0.beta1) lib/active_support/callbacks.rb:636:in `skip_callback'
[...]

PR #496, opened 2016-01-15
Remove exception from skip_callback

jeryRazakarison:
skip_callback will raise an exception if the callback is not set by Devise.
But Devise doesn't always set the callback `send_on_create_confirmation_instructions`, (for example when :confirmable is not present).
In several cases, with the master branch of Devise, we then have an error 500
```
Completed 500 Internal Server Error in 78ms (ActiveRecord: 0.0ms)
ArgumentError (After create callback :send_on_create_confirmation_instructions has not been defined):
...
/Users/jery/.rvm/gems/ruby-2.2.2/bundler/gems/devise_token_auth-d154b0294006/app/controllers/devise_token_auth/registrations_controller.rb:39:in `create'
```
Adding the option :raise to false (as stated in rails documentation), it doesn't raise an exception when the callback has not been set by Devise, and fixes issue #397

### DIFF
--- a/app/controllers/devise_token_auth/registrations_controller.rb
+++ b/app/controllers/devise_token_auth/registrations_controller.rb
@@ -36,7 +36,7 @@ module DeviseTokenAuth
 
       begin
         # override email confirmation, must be sent manually from ctrl
-        resource_class.skip_callback("create", :after, :send_on_create_confirmation_instructions)
+        resource_class.skip_callback("create", :after, :send_on_create_confirmation_instructions, raise: false)
         if @resource.save
           yield @resource if block_given?
 


britg:
@jeryRazakarison I'm running into this issue. Why did you close this PR? I can't find any further discussion on the correct way to do this.

britg:
I ended up creating an override as below, but it seems an unnecessary hack?

````ruby
class Overrides::RegistrationsController < DeviseTokenAuth::RegistrationsController

  def create
    resource_class.set_callback("create", :after, :send_on_create_confirmation_instructions)
    super
  end

end
````

jeryRazakarison:
Hi @britg,
I closed this PR because it didn't pass the Travis Build, and so there was a risk it created unwanted side effects for other users.
I ended up overriding the `create` method following these instructions https://github.com/lynndylanhurley/devise_token_auth#custom-controller-overrides, but using
`resource_class.skip_callback("create", :after, :send_on_create_confirmation_instructions, raise: false)`
instead of
`resource_class.skip_callback("create", :after, :send_on_create_confirmation_instructions)`

I couldn't find a correction that passed the Travis CI build.

(Overriding was very efficient in my case, because I was able to make devise_token_auth works well with devise_invitable.)

britg:
@jeryRazakarison Ok, thanks for the explanation. Looks like overriding is the way to go for now.

PR #510, opened 2016-01-23
Allow backup identification for account finding

### LINKED ISSUE #509: Allow backup field for login
The most common case of this is having:
 + a user model with an email
 + a profile model with a username

I want a way to easily add an alternate login method (i.e. a profile's username)

----------------------

similar to:

 + #53 
 + #127 
 + #128 
 + #395 
 + #485

djsegal:
Attempt to resolve #509

Do you have any insight on how to test this?

---------

**parameters:**
 + `backup_field_name`: something like `username`
 + `backup_field_class`: something like `user` or `profile` (defaults to `resource_class`)
     - if another class is used, require it to have a direct 1-to-1 association with `resource_class`
     - i.e. `profile.user` must return the user of a profile

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -12,25 +12,19 @@ module DeviseTokenAuth
       # Check
       field = (resource_params.keys.map(&:to_sym) & resource_class.authentication_keys).first
 
-      @resource = nil
-      if field
-        q_value = resource_params[field]
-
-        if resource_class.case_insensitive_keys.include?(field)
-          q_value.downcase!
-        end
-
-        q = "#{field.to_s} = ? AND provider='email'"
-
-        if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
-          q = "BINARY " + q
-        end
-
-        @resource = resource_class.where(q, q_value).first
+      q_value = resource_params[field]
+      if resource_class.case_insensitive_keys.include?(field)
+        q_value.downcase!
       end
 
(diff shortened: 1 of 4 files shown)


app/controllers/devise_token_auth/sessions_controller.rb
```diff
@@ -12,25 +12,19 @@ def create
...
-          q = "BINARY " + q
-        end
+      q_value = resource_params[field]
+      if resource_class.case_insensitive_keys.include?(field)
+        q_value.downcase!
+      end
 
-        @resource = resource_class.where(q, q_value).first
+      set_resource(field, q_value)
+      unless @resource.present?
```
djsegal: Anything below here probably isnt needed (besides the `def resource_params...end` block)

test/dummy/app/controllers/overrides/sessions_controller.rb
```diff
@@ -5,7 +5,7 @@ class SessionsController < DeviseTokenAuth::SessionsController
     def create
       @resource = resource_class.find_by_email(resource_params[:email])
 
-      if @resource and valid_params?(:email, resource_params[:email]) and @resource.valid_password?(resource_params[:password]) and @resource.confirmed?
+      if @resource and resource_params[:password].present? and @resource.valid_password?(resource_params[:password]) and @resource.confirmed?
```
djsegal: If `valid_params?` wasn't changed in `sessions_controller`, this wouldn't be needed

djsegal:
@lynndylanhurley i know youve wanted to keep this sort of logic inside devise, but I thought it might be useful for people at this level.

djsegal:
Made obsolete by #538

PR #519, opened 2016-01-31
allow only one confirmation

rmvenancio:
When a user confirms and account (after a registration), if the registration is valid, the user is logged in automatically in the system. 

After a successful confirmation the user can still use the confirmation email to login into the system (without a password) and this represents a security flaw. 

If the token is valid and if the user is already confirmed, the devise_token_auth should return an expired token. This will raise an auth:email-confirmation-error in ng-token-auth. 

This will require #410 because it looks for sign_in counter.

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -10,11 +10,16 @@ module DeviseTokenAuth
         token_hash = BCrypt::Password.create(token)
         expiry     = (Time.now + DeviseTokenAuth.token_lifespan).to_i
 
+        if @resource.sign_in_count >0 
+          expiry     = (Time.now + 1.second).to_i
+        end
+        
         @resource.tokens[client_id] = {
           token:  token_hash,
           expiry: expiry
         }
 
+        sign_in(@resource)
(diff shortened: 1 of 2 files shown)


zachfeldman:
@lynndylanhurley appears to be a decently sized security hole that's also easy to merge!

zachfeldman:
I'm down to merge this right after https://github.com/lynndylanhurley/devise_token_auth/pull/410 if it gets approval from one other person.

lynndylanhurley:
Changes look good, I'll resolve the conflicts with the test file. We can merge once the conflicts are resolved.

ethagnawl:
It's very encouraging that each issue I've bumped while using this library has either been 1) recently fixed, 2) has a PR pending or 3) has a documented workaround.

👍

rmvenancio:
@maicolben Hi. I Don't have a Dev environment anymore. I can't help at this time.

MaicolBen:
No problem, when I have a chance I'll do it myself

rmvenancio:
Thanks for the understanding. Continue with the good work. BR.

MaicolBen:
I opened this PR with the conflicts solved (because I can't push to a fork)

PR #523, opened 2016-02-04
create token when no client_id token

### LINKED ISSUE #286: NoMethodError (undefined method `[]' for nil:NilClass):
I have this error for a couple of day now. Any leads to solve ti ?

```
NoMethodError (undefined method `[]' for nil:NilClass):
  devise_token_auth (0.1.32.beta10) app/models/devise_token_auth/concerns/user.rb:180:in `build_auth_header'
  devise_token_auth (0.1.32.beta10) app/controllers/devise_token_auth/concerns/set_user_by_token.rb:68:in `update_auth_header'
````

charlesdg:
fix #286. Case of no change headers on each request.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -196,18 +196,23 @@ module DeviseTokenAuth::Concerns::User
 
   def build_auth_header(token, client_id='default')
     client_id ||= 'default'
+    
+    if !DeviseTokenAuth.change_headers_on_each_request && self.tokens[client_id].nil?
+      create_new_auth_token(client_id)
+    else
 
-    # client may use expiry to prevent validation request if expired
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
@charlesdg - could you provide any background on what you think the underlying issue is with `client_id` being unavailable? Going to merge since this adds some resiliency, but would love to understand the root issue.

hiattp:
@booleanbetrayal as I mentioned in the issue comments this seemed fairly clearly to be a concurrency issue, best solved by locking the user record while updating the auth hash

stevehanson:
I just confirmed that this fixed the issue for me, but it doesn't look like there's been a release since this was merged. Is this planned to go live in the near future? 

In the mean-time, I'll just load the gem from GitHub using the latest master commit.

PR #525, opened 2016-02-05
Extract model callbacks

merqlove:
Added these changes for the case if we need custom callbacks. 
For example, i have User model with multiple providers, so model don't have fields provider & uid at all, instead it `has_many` on `oauths`.

* Extract model callbacks. 
* Add setting for extracted callbacks. Load it by default. 
* Update generator template. 
* Add line to readme.


booleanbetrayal:
LGTM! Thanks @merqlove !

PR #528, opened 2016-02-08
Added omniauth post route

v3rtx:
Some of the omniauth providers such as steam response with POST message which we also want to handle.


v3rtx:
I've faced with the problem that steam send POST message as response result of the OAuth, in earlier versions there was route for post so I guess this fix is ok. Please feel free to contact me if you have any notes/questions.

v3rtx:
@lynndylanhurley , @booleanbetrayal Hey guys! What is need to be done in this PR before merging it?

booleanbetrayal:
thanks @v3rtx ... merging, but if you could post a list of providers you've encountered that require POST, it'd be nice to document.

v3rtx:
@booleanbetrayal Thanks! I'm personally use Steam OAuth (https://steamcommunity.com/) and it sends post request in response. All other OAuth providers which I use works with get.

PR #531, opened 2016-02-09
Avoid sending auth headers if while processing used token is cleared

virginia-rodriguez:
This pull request fix the error occurred in following scenario:

1. Client sends request R1 and server starts processing it
2. Client sends sign out request R2 and server process it (R1 still processing)
3. R1 finishes. Its response to the server will contain auth headers, although user already signed out.

To avoid R1 response containing auth headers, in the after action code, it should be checked that the token had not been removed in the meantime.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -74,6 +74,10 @@ module DeviseTokenAuth::Concerns::SetUserByToken
     @client_id = nil unless @used_auth_by_token
 
     if @used_auth_by_token and not DeviseTokenAuth.change_headers_on_each_request
+      # should not append auth header if @resource related token was
+      # cleared by sign out in the meantime
+      return if @resource.reload.tokens[@client_id].nil?
+
       auth_header = @resource.build_auth_header(@token, @client_id)
 
       # update the response header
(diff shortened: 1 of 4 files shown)


app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -74,6 +74,10 @@ def update_auth_header
     @client_id = nil unless @used_auth_by_token
 
     if @used_auth_by_token and not DeviseTokenAuth.change_headers_on_each_request
+      # should not append auth header if @resource related token was
+      # cleared by sign out in the meantime
+      return if @resource.reload.tokens[@client_id].nil?
```
booleanbetrayal: @virginia-rodriguez - any implementation ideas that don't involve the additional `reload` call? i'm personally aiming for sub-200ms roundtrip responses in my apps these days and that's pretty hard to do with Rails and additional ActiveRecord calls.  =]

booleanbetrayal:
Thanks for this PR @virginia-rodriguez . I had one inline comment. I'm wondering if we can avoid some DB overhead. If we can't think of something incredibly creative soon, I'll go ahead and merge it.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
virginia-rodriguez: @booleanbetrayal I can't think of any way to avoid the reload. In fact, when DeviseTokenAuth.change_headers_on_each_request is true (default value, right?) an implicit reload is already done by using @resource.with_lock

booleanbetrayal: Yeah ... me neither. was trying to avoid that call when `change_headers_on_request` is `false`. Might need to revisit optimistic locking in general. Couple of other related issues. Merging and may revisit later!

PR #557, opened 2016-03-02
@resource.save! to have 400 error if validation failed.

charlesdg:
add a ! to the save to raise an ActiveRecord::RecordInvalid exception to have a 4*\* http code in the response if the validation failed.


zachfeldman:
@lynndylanhurley possibly easy to merge...unsure if it's wise though

MaicolBen:
How can a model be invalid only saving the tokens? Did you add a validation to the model related to the tokens

zachfeldman:
@MaicolBen this is....also a very old PR. If you think it doesn't make much sense anymore, feel free to close it. @charlesdg we welcome your opinion as well.

PR #561, opened 2016-03-04
DeviseTokenAuth.enable_standard_devise_support changes

mgwidmann:
Calling `@resource.create_new_auth_token` simply ends up revoking older tokens on every request since the web (legacy devise) ends up using this part of the code.

Also, we ran into problems with the `sign_in` call not using the `mapping` variable for the first parameter. Additionally, a devise bug when using both `store: false` and `bypass: true` (https://github.com/plataformatec/devise/issues/3981) causes the user object to be stored in the session which will not be used since it was an API request to begin with. This bloats our redis memory usage significantly (we have a large user model).

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -37,7 +37,6 @@ module DeviseTokenAuth::Concerns::SetUserByToken
       if devise_warden_user && devise_warden_user.tokens[@client_id].nil?
         @used_auth_by_token = false
         @resource = devise_warden_user
-        @resource.create_new_auth_token
       end
     end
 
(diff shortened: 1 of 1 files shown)


mgwidmann:
Waiting to hear back before fixing tests.

mgwidmann:
@lynndylanhurley ping, hoping to hear your thoughts on this.

jphager2:
I also came across the issue with the `:user` mapping used, but I was able to get around that (using `devise_auth_token_group :user, contains: [:my_mapping]`

What was a bigger issue for me is that the mapping passed to `Warden::Proxy#user` is not the argument `mapping` passed to `set_user_by_token`, but instead a underscored symbol of the `resource_class`.  In 1.3.7, we avoid this by setting `DeviseTokenAuth.enable_standard_devise_support` to `false`, but I'm not sure that this is desirable.

mgwidmann:
looks like no one cares about this repo and this project should just die and be deleted...

PR #537, opened 2016-02-12
Make all json responses to be json_api compliant

djsegal:
Resolves #536

PR #562, opened 2016-03-06
Return resource with top-level 'type' member.

ruimiguelsantos:
Render resource with top-level 'type' member on responses from SessionsController
and TokenValidationsController for json:api compliance.

Just an extension of the work done on #537 .


booleanbetrayal:
Thanks @ruimiguelsantos !

PR #564, opened 2016-03-08
User concern: Ensure fallback is in place

olleolleolle:
Without this change "redirect_url" may be blank, leading to an internal bad URI error unless set when confirming.

Occurs when creating a user from the Rails console, for instance.

**Workaround**: implement the `#send_confirmation_instructions` method in your User model, with the `opts[:redirect_url] ||= DeviseTokenAuth.default_confirm_success_url` line in there.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -70,6 +70,7 @@ module DeviseTokenAuth::Concerns::User
       if pending_reconfirmation?
         opts[:to] = unconfirmed_email
       end
+      opts[:redirect_url] ||= DeviseTokenAuth.default_confirm_success_url
 
       send_devise_notification(:confirmation_instructions, @raw_confirmation_token, opts)
     end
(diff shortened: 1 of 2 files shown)


booleanbetrayal:
Looks sane to me @olleolleolle . Could you please fix the tests in an amended commit?

olleolleolle:
The tests failed because the fixes to the regexps that had been introduced in the "normal User model" section of the file was not shared to the "alt User model" section.

I'm weak on the style of minitest methods, but perhaps a shared method `token_and_redirect_url_from(body)` would help?

olleolleolle:
Now, with a shared test method.

olleolleolle:
@booleanbetrayal Rereview?

booleanbetrayal:
thanks @olleolleolle and sorry for the delay!

olleolleolle:
Cool! Thanks for building this project! I use it and it does so many things for me. Thanks for brightening my day.

PR #566, opened 2016-03-08
JSON API v1.0 response formatting support

Dschee:
To make it short the goal of this PR is to **add support for rendering all API responses in the JSON API specification format (v1.0)**. See the official documentation [here](http://jsonapi.org/format/) for more information.

My plan is to (as @booleanbetrayal suggested):
- [x] Restructure code keeping _backwards compatibility_ to the existing custom style
- [x] Add JSON API support as a _configurable option_ (disables JSON API responses entirely)
- [x] _Use the JSON API formatting_ when HTTP_HEADER is set to `application/vnd.api+json`

Introduce new JSON API compliant responses ([Reference](http://jsonapi.org/format/))
- [x] Passwords
- [x] Registrations
- [x] Sessions
- [x] Token Validations

_Update specs_ to cover the new JSON API response formatting
- [x] Passwords 
- [x] Registrations
- [x] Sessions
- [x] Token Validations
- [x] _Document_ the usage (via README) and implementation details (via comments)

The following are knowingly **excluded from this PR** (can be added separately):
- _Deserialize_ JSON API compliant formatted requests
- _Update specs_ to cover the new JSON API request formatting

See #596 for a start of these.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gemspec
 
 group :development, :test do
   gem 'thor'
-  gem "figaro",                 :github => 'laserlemon/figaro'
+  gem 'figaro',                 :github => 'laserlemon/figaro'
   gem 'omniauth-github',        :git => 'git://github.com/intridea/omniauth-github.git'
   gem 'omniauth-facebook',      :git => 'git://github.com/mkdynamic/omniauth-facebook.git'
   gem 'omniauth-google-oauth2', :git => 'git://github.com/zquestz/omniauth-google-oauth2.git'
@@ -24,8 +24,8 @@ group :development, :test do
 
   # testing
   #gem 'spring'
-  gem "pry"
-  gem "pry-remote"
+  gem 'pry'
+  gem 'pry-remote'
   gem 'minitest'
   gem 'minitest-rails'
   gem 'minitest-focus'
@@ -40,9 +40,9 @@ end
 
 # code coverage, metrics
 group :test do
-  gem "codeclimate-test-reporter", require: nil
+  gem 'codeclimate-test-reporter', require: nil
 end
 
 group :development do
-  gem "github_changelog_generator"
+  gem 'github_changelog_generator'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ group :development, :test do
   gem 'faker'
   gem 'fuzz_ball'
   gem 'mocha'
+  gem 'json_expressions'
 end
 
 # code coverage, metrics
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -229,7 +229,7 @@ GEM
       railties (>= 4.2.0, < 5.1)
     ruby-progressbar (1.7.5)
     shellany (0.0.1)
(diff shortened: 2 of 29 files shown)


Dschee:
@djsegal I just came across the `is_json_api` method in `application_controller` and wonder what it is for. My question especially is if a JSON API implementation does interfere in any way with that method – do we need to update it once JSON API responses are possible? Or is it for something completely different?

Would be great to get some insights.

Dschee:
I have now completed the JSON API responses. I opted to do this first before beginning the work on JSON API compliant requests and I feel we can merge the two parts independently as the HTTP header fields `ACCEPT` and `CONTEN_TYPE` are separate, too.

This implementation doesn't change the behavior of the current custom format at all, simply adding JSON API responses if a client requests them by specifying the `ACCEPT` HTTP header as `application/vnd.api+json`.

I also did a few smaller changes (like fixing clearly machine translated German localizations, using single quotes instead of double quotes in some places). Those should all be okay, I think.

Could you please review this PR and merge it, if you think it's okay? @booleanbetrayal

Dschee:
Any thoughts on merging this? @booleanbetrayal @lynndylanhurley

djsegal:
@Dschee my bad about not responding.

`is_json_api` is just a function to robustly test if the user is using json api (i.e. it doesn't break on rails 4)

djsegal:
@Dschee also, it might be good to rebase/refactor this PR to increase readability.

for example, like 90% of the line differences are in tests

and all the `case response_format` things might be able to be refactored out so we don't triple the size of each controller

ValentinTrinque:
I agree with @djsegal. I think we can refactor this a bit. What about implementing Responders objects, let's say `JsonApiResponder` and `JsonResponder`, holding the `render` logic of their own ?

Then, instead of using a conditional statement to set `response_format` we can directly set the responders to call, since they will have the same API.

**Example :**

```ruby
def responder
  case expected_format
  when :json_api
    JsonApiResponder.new
  when :json
    JsonResponder.new
  else
    NullResponder.new
  end
end


if @errors
  return responder.render_create_error_missing_redirect_url(args)
end
```

This is a conceptual piece of code. The

We could also have a class to unify the I18n calls. Having to duplicate the I18n key across the different `render` calls seems error prone to me.

Dschee:
@ValentinTrinque Good idea actually, I won't have time to do it myself though, I moved on from this gem long time ago as a reaction to my suggestion took several months. But I'd say your approach is much cleaner than what I have started. Would love to see this in dta sometime.

ValentinTrinque:
@Dschee Oh wow, I didn't look at the dates... I understand your feeling then... I am not really willing to contribute if there is no feedback neither.

djsegal:
@Dschee & @ValentinTrinque: 
+ I don't want this to be how this gem dies.
+ I'll join the guard and help work through this if needed.

zachfeldman:
Sorry about that @Dschee we appreciate your contribution. @ValentinTrinque or anyone else, if you ever want to pick this back up, feel free to reopen this PR.

PR #569, opened 2016-03-11
now possible to change headers names in the config file

ingolfured:
Would be very handy to change the headers names because, for example, `expiry` conflicts with another expiry which is coming from a different project.


booleanbetrayal:
LGTM!

PR #576, opened 2016-03-16
allowing authenticating using headers as well as a post request

ingolfured:
For some system, all post request data is logged in the system. When there is no possible way to change that, it's much nicer to be able to send the user email and password via the header instead of a data string in a post request. Therefore, instead of sending

```bash
curl -s -X POST -d 'email=some@cool.com&password=partypassword' -D -  'https://theurl.com/api/v1/auth/sign_in' -o /dev/null 2>&1
```

we can do

```bash
curl -s -X POST -H 'email: some@cool.com' -H 'password: partypassword' -D -  'https://theurl.com/api/v1/auth/sign_in' -o /dev/null 2>&1
```

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -17,6 +17,9 @@ module DeviseTokenAuth
     protected
 
     def params_for_resource(resource)
+      devise_parameter_sanitizer.instance_values['permitted'][resource].each do |type|
+        params[type.to_s] ||= request.headers[type.to_s] unless request.headers[type.to_s].nil?
+      end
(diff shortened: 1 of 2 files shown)


ingolfured:
You guys have any thoughts?

lynndylanhurley:
Hi @ingolfured - this looks fine to me.

lynndylanhurley:
But if we're going to support this, should it work the same way for each action (registration, password reset, etc.)?

ingolfured:
Pretty sure this handles the other controllers as well (all calling `params_for_resource`.) Are you talking about writing more tests to cover that? Or am I misunderstanding something?

PR #589, opened 2016-03-29
Update readme for headers names

y4ashida:
Ref: #569

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,9 +325,15 @@ module YourApp
       allow do
         origins '*'
         resource '*',
-          :headers => :any,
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
thanks @y4ashida !

y4ashida:
@booleanbetrayal 
I realized this commit isn't good, because 'config/applicaiton.rb' is loaded before `config/initializers/devise_token_auth`.
So, if you change headers names in `config/initializers/devise_token_auth`, the changes isn't reflected in 'config/applicaiton.rb'.

Moreover, `DeviseTokenAuth.headers_names` isn't defined in some cases. #591

Sorry to trouble you. I think this commit should be reverted.

PR #608, opened 2016-04-13
if token is nil, create new_auth_token

charlesdg:


booleanbetrayal:
I don't understand the use case in which `token` would be provided as `nil`, but it's probably best wrapping a conditional around that, assuming it's legitimate.

charlesdg:
When we ask for a password recovering, we already have a token for the client_id in the database but the user doesn't have it because he came from an email and he's not authenticated. So we need, a token for him.

By wrapping a conditional around that, what do you mean?

PR #612, opened 2016-04-17
Modified condition to apply unique_email_user validation

salmanasiddiqui:
`before_create :sync_uid` is not needed as `before_save :sync_uid` is already defined.

Moved conditions out from `unique_email_user` and `sync_uid`, so that anyone overriding these methods wouldn't have to take care of those conditions.

Updated condition on which to apply `unique_email_user` validation. It was only being validated _on create_, while it should be applied whenever _email is changed_.


zachfeldman:
Hi there @salmanasiddiqui    ,

Hope all is well.

salmanasiddiqui:
hey @zachfeldman , although this isn't a big PR, but i think its better if we merge these changes in.
By moving away the conditions from the function onto the callbacks, it makes it easier for others to overwrite those functions without copying over those conditions.

lynndylanhurley:
@salmanasiddiqui I agree that this PR is an improvement. Can you please provide tests?

PR #638, opened 2016-05-17
Fix yields from controller actions

tiagojsag:
So let me start by saying that this is my first PR to a ruby gem. So, if it's useless/crap, feel free to tell me so (gently).

The problem comes from the following example: https://github.com/lynndylanhurley/devise_token_auth#passing-blocks-to-controllers

Currently only the registration controller seems to be yielding properly (that is, something at all). All other controller actions seem to not yield anything.
I added the tests to back my theory up, along with the fairly simple fixes.

Heads up to session/destroy action, which is yielding 'user', not '@resource'.

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -17,7 +17,7 @@ module DeviseTokenAuth
 
         @resource.save!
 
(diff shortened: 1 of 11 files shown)


booleanbetrayal:
Seems more useful to anyone extending these guys, for sure. Thanks @tiagojsag !

PR #640, opened 2016-05-17
Updating Devise dependency to max 4.1.1

TGRGIT:
Updating devise dependency to 4.1.1 from 4.1


TGRGIT:
Abondoning, will try without updating gempec

PR #648, opened 2016-05-27
updates config wrapper to conform with newer idiom

bvandgrift:


booleanbetrayal:
thanks @bvandgrift !

bf4:
@bvandgrift which newer idiom is this supposed to address?  The newest AMS doesn't have this method.  And if the idiom is Ruby, it's kind of weird to use the return result of a configure block that has non-config logic in it...

PR #650, opened 2016-06-02
Tokens count overmuch fixed

JerryGreen:
When we set `config.change_headers_on_each_request` to `false` it is possible to get an unlimited count of tokens and `config.max_number_of_devices` is ignoring.

In a casual case - it's not a problem because we remove the token on sign_out. But if your application is an API - nobody uses sign_out and you get over 19k tokens for a single object or so (for example).

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -177,6 +177,17 @@ module DeviseTokenAuth::Concerns::User
       updated_at: Time.now
     }
 
+    return build_auth_header(token, client_id)
+  end
+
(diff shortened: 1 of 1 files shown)


booleanbetrayal:
@JerryGreen - tests are failing on `bundle install`. can you rebase this PR branch to exclude the Gemfile change? I'd definitely take a dependency update with passing tests in a separate PR if it helps.

JerryGreen:
Did it.

Why don't you describe a certain version for every dependency?

booleanbetrayal:
thanks @JerryGreen !

Issue #644, opened 2016-05-19
Issues with active_model_serializers


virginia-rodriguez:
I was using active_model_serializers 0.9.5 with last released version of devise_token_auth (0.1.37) and everything worked fine. Then I tried to use devise_token_auth master branch to use a fix that was present there. But other changes introduced to master branch cause devise_token_auth to stop working properly (ex. sign in is broken). An exception is raised in https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/application_controller.rb#L38, because ActiveModel::Serializer config is not defined.
`NoMethodError (undefined method 'config' for ActiveModel::Serializer:Class):
`

Also, since devise_token_auth seems to be having some issues with last version of active_model_serializers (see https://github.com/lynndylanhurley/devise_token_auth/issues/600), why force to use last version? 

What about just replacing the conflictive line with:
`ActiveModel::Serializer.respond_to?(:config) && ActiveModel::Serializer.config.adapter == :json_api`

Anyone is having similar issues?

SarSar440:
Not sure if this is related but I have since updated my Active model serializer gem and suddenly the client authentication header is now default.

SebAshton:
@virginia-rodriguez I had this issue, and solved it by specifying version `'~> 0.10.0'` in my Gemfile.

e.g. `gem 'active_model_serializers', '~> 0.10.0'`

augustosamame:
I am getting a similar error on sign-in / sign-up:

NoMethodError (undefined method `setup' for ActiveModel::Serializer:Class):
  
/home/augusto/.rvm/gems/ruby-2.3.0@faces/bundler/gems/devise_token_auth-81fc7b40998d/app/controllers/devise_token_auth/application_controller.rb:38:in `is_json_api'

I have the following in my gem file:

gem 'devise'
gem 'devise_token_auth', :git => 'git://github.com/lynndylanhurley/devise_token_auth.git'
gem 'active_model_serializers', '~> 0.10.0'

Any ideas?

augustosamame:
This https://github.com/lynndylanhurley/devise_token_auth/commit/7b9727995e1eb42385a2a4102516fcd3a372ad81 commit is not compatible with current 0.10.0 version of Active Model Serializer. Patching now

booleanbetrayal:
i'll take that PR whenever you get a chance @augustosamame -- the state of things in AMS is kinda confusing these days.

[...]

PR #651, opened 2016-06-02
fix method 'is_json_api' with active_model_serialier v 0.10.0

woodcrust:

PR #653, opened 2016-06-02
updates config wrapper to conform with multiple versions of Active Model Serializers

augustosamame:
Fixes config adapter wrapper so it will be compatible with multiple versions of Active Model Serializer. Tested with 0.9.2 and 0.10.0 versions of AMS

See Issue #644


booleanbetrayal:
Thanks @augustosamame, but this check actually went in with #651 just a moment ago.

booleanbetrayal:
PS - I actually thought that was your PR when it came in, mobile device GitHub fail. Thanks for helping shine some light on this issue.

PR #673, opened 2016-07-08
Respect relative_url_root

okuramasafumi:
It didn't work properly when Rails works in a subdirectory, so I made it respect `Rails.application.config.relative_url_root`.

### DIFF
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -13,7 +13,8 @@ module DeviseTokenAuth
       # before authentication.
       devise_mapping = [request.env['omniauth.params']['namespace_name'],
                         request.env['omniauth.params']['resource_class'].underscore.gsub('/', '_')].compact.join('_')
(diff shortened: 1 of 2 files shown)


app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
```diff
@@ -13,7 +13,7 @@ def redirect_callbacks
       # before authentication.
       devise_mapping = [request.env['omniauth.params']['namespace_name'],
                         request.env['omniauth.params']['resource_class'].underscore.gsub('/', '_')].compact.join('_')
-      redirect_route = "#{request.protocol}#{request.host_with_port}/#{Devise.mappings[devise_mapping.to_sym].fullpath}/#{params[:provider]}/callback"
+      redirect_route = "#{request.protocol}#{request.host_with_port}/#{Rails.application.config.relative_url_root}/#{Devise.mappings[devise_mapping.to_sym].fullpath}/#{params[:provider]}/callback"
```
booleanbetrayal: @okuramasafumi - shouldn't we be doing the same `relative_url_root.present?` check here, as well, so as not to generate double-slashes?

PR #697, opened 2016-08-13
add zh-CN.yml

halfray:


booleanbetrayal:
Thanks for this @halfray !

Issue #698, opened 2016-08-14
Deprecation Error Message on 5.0


juddey:
Just ran a migration using the generated template and got this deprecation notice:

``` 
DEPRECATION WARNING: Directly inheriting from ActiveRecord::Migration is deprecated. Please specify the Rails release the migration was written for:

class DeviseTokenAuthCreateUsers < ActiveRecord::Migration[4.2] (called from <main> at bin/rake:4)
```
Let me know if you want a PR to fix and I'll cook something up.

lynndylanhurley:
>Let me know if you want a PR to fix and I'll cook something up.

Yes please 👍

PR #700, opened 2016-08-14
Fix Migration Deprecation Warning

juddey:
Fix for issue #698 
Update migration template to include version of Rails to silence a Rails 5.0 deprecation warning.

Introduced new logic to in the generator template and added a test to ensure the version number was written to the migration file.

Great to be able to give back 😄


booleanbetrayal:
thanks @juddey

PR #703, opened 2016-08-16
Always set header in batch mode

### LINKED ISSUE #702: Always Set Headers in Batch Mode
The response headers are only set if we are not in batch request:

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L121

This causes issues with Rails' Etag support and XHR Requests: If there are no token_auth headers and Rails returns a 304 Not Modified, the browser (Chrome) will use a cached response, including all headers of the cached response (access-token, client, ...). There is no way in the client to distinguish these cached headers from server-set ones. Eventually, a cached client id is stored and used in the subsequent request, resulting in a 401.

I moved the linked line (121) after the if-block, causing the headers always to be set, then everything worked as expected.

codez:
Fixes #702

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -113,13 +113,21 @@ module DeviseTokenAuth::Concerns::SetUserByToken
         if @is_batch_request
           auth_header = @resource.extend_batch_buffer(@token, @client_id)
 
+          # Do not return token for batch requests to avoid invalidated
+          # tokens returned to the client in case of race conditions.
+          # Use a blank string for the header to still be present and
+          # being passed in a XHR response in case of
+          # 304 Not Modified responses.
+          auth_header[DeviseTokenAuth.headers_names[:"access-token"]] = ' '
+          auth_header[DeviseTokenAuth.headers_names[:"expiry"]] = ' '
+
         # update Authorization response header with new token
         else
           auth_header = @resource.create_new_auth_token(@client_id)
-
-          # update the response header
-          response.headers.merge!(auth_header)
         end
+
+        # update the response header
+        response.headers.merge!(auth_header)
 
       end # end lock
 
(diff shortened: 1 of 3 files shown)


lynndylanhurley:
The client libraries save new token information that is returned with each response.

Will this PR require modifications to the client libraries?

beantoan:
Both of changing header on each request and batch mode, client should receive auth headers in response.

ryaz:
Hi @lynndylanhurley!
Did you have chance to look into it?

bernabe9:
Hi @lynndylanhurley, any news about this PR?

bartes:
+1 for merging this

nijikon:
Can we merge this please? And my other question is what is the status of this, there are plenty of PRs open, do you need some help on this repository to handle all that PRs and issues?

booleanbetrayal:
Is anyone running this with ng-token-auth atm? Since these libraries are frequently paired, I don't want to break the response handling. Can someone confirm?

bartes:
@booleanbetrayal yes  with ng-token-auth

zachfeldman:
@nijikon we'd love some help to handle PRs and issues - @bartes as well! If you have time to review open PRs and tell us if you think they're good to go, feel free to respond here :)

zachfeldman:
Changes look good to me and address a long standing issue. All checks are passing and approved by @MaicolBen ...going to merge this one

krzysiek1507:
I'm not sure if it was just a bugfix, but it introduced a breaking change and IMO it shouldn't be released as a patch version. Unfortunately, `redux-auth` doesn't work with `v0.1.43` and @lynndylanhurley is not active anymore so it's very bad.

krzysiek1507:
Maybe we should rollback this, release `v1.0.0`, branch `v1` and follow semver? Breaking changes should be only in major releases. It would be also great to have smaller releases and more often (e.g. one per month) instead of big ones. What do you think?

MaicolBen:
@krzysiek1507 Agreed, but I am not in charge of releasing the gem. I asked Lynn, and we agreed to [rollback this](https://github.com/lynndylanhurley/devise_token_auth/pull/1161) and release 0.2.0. The next release will be 1.0.0

MaicolBen:
0.2.0 was released with this rollback, and 1.0.0rc1 is released with the current master, sorry for the delay!

PR #704, opened 2016-08-16
Allow user specific token lifespans

codez:
Instead of keeping the token lifespan only in the global configuration, we added a corresponding method in the User Module. This method may be overriden to return a user specific value for the token lifespan. By default, the global value is used.

Use case: We have users working all day with the application in the company headquarters, who do not want to login after short breaks. Other users access the application just for a few minutes from public computers, for whom we prefer to have a very short token lifespan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,7 +484,7 @@ Models that include the `DeviseTokenAuth::Concerns::User` concern will have acce
   # store client + token in user's token hash
   @resource.tokens[client_id] = {
     token: BCrypt::Password.create(token),
-    expiry: (Time.now + DeviseTokenAuth.token_lifespan).to_i
+    expiry: (Time.now + @resource.token_lifespan).to_i
   }
 
   # generate auth headers for response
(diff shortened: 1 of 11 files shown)


MaicolBen:
Good work, but seems a specific use case of this gem, I don't know if everyone wants that. Don't you need a migration for this?

codez:
That's why I implemented the default [token_lifespan](https://github.com/lynndylanhurley/devise_token_auth/pull/704/files#diff-7721c6c51a44a791536d9eb0cbe4ea4cR233) method of a `User` it in an unobtrusive way. The behavior remains the same if you do not want to use this feature. You do not have to care in this case.

If the feature is used, it is left to the programmer to provide a decent implementation for this method. The most obvious case is an accessor method for a database column, and the a migration would be needed. But it really might be anything, so the programmer has full flexibility.

zachfeldman:
Seems fairly straightforward, especially if it doesn't impact default behavior.

MaicolBen:
Yeah, this gives more flexibility. I checked and `resource` isn't null everywhere you replaced the default behavior.

zachfeldman:
👍 we should cut a release soon...

PR #722, opened 2016-08-31
Memorize current_#{group_name} to avoid error

Charlie-Hua:
Currently `current_#{group_name}` (e.g. `current_member`) is **NOT** memorized as `current_#{mapping}` (e.g.`current_user`) is.

Instead, it calls `set_user_by_token` **every time**, and this could cause error if `current_member` is called after token is changed (maybe by some other request) and return nil.

This PR also removes unnecessary `set_user_by_token` from `authenticate_member!`


zachfeldman:
Seems easy to review and possible merge, if it makes sense. It's a short one

PR #724, opened 2016-08-31
Update indexes on template

dijonkitchen:
Why:
Keep data consistent with multiple processes, match devise gem implementation, and latest ruby hash syntax convention.

References:
https://robots.thoughtbot.com/the-perils-of-uniqueness-validations
https://github.com/plataformatec/devise/blob/master/lib/generators/active_record/templates/migration.rb#L13


booleanbetrayal:
Sounds good if it's consistent with devise itself. Thanks @dijonkitchen

PR #730, opened 2016-09-10
Adds method to builder resource

rodrigoulisses:
This method build the resource and is necessary to developer set the
others attributes of resource.


zachfeldman:
@rodrigoulisses closing this - I don't really see exactly how this helps. Glad to hear a justification/reopen if you disagree.

PR #737, opened 2016-09-19
Conditionally set rails version on migration

akz92:
I added a condition that only appends the Rails version to the migration if it's 5 or greater (when this syntax was added). This code was causing issues when running the migration on Rails 4.


zachfeldman:
Makes sense to me but the build fails. @akz92 could you look into this? We will close in 7 days if we don't hear from you but just let us know if you need more time!

MaicolBen:
@akz92 I reproduced the issue and your PR solves it. I believe travis failing due to an old build that was wrongly merged. Can you rebase it? Otherwise if we don't hear from you in a week. I will cherry-pick this PR and do it myself.

akz92:
@MaicolBen I rebased it but apparently something is still wrong. It's been a long time since I created the pull request, so it would take me some time to figure this out =/.

MaicolBen:
Weird, maybe because you did it from master. I guess you should force push your master with lastest code of this gem and add your commit

MaicolBen:
I recreated the PR #979

PR #746, opened 2016-10-05
Use standart ActiveRecord error message for email uniqueness validation

mpugach:
According to the [Rails Guides](http://guides.rubyonrails.org/v5.0/i18n.html#error-message-interpolation) there already is a localization for the uniqueness validation. And the key is `:taken`.

So if one wants to provide his own translation, he should redefine one of the keys in the following order:

```
activerecord.errors.models.user.attributes.email.taken
activerecord.errors.models.user.taken
activerecord.errors.messages.taken
errors.attributes.email.taken
errors.messages.taken
```


booleanbetrayal:
LGTM!

booleanbetrayal:
thanks @mpugach

PR #755, opened 2016-10-31
change auth_token to access-token and client_id to client in params 

bonyiii:
I think these parameter names needs to be changed. After this change the application was able display omniauth params after redirect.


bonyiii:
Hi @zachfeldman, 

I'm not sure whether it was real a bug and I haven't deal with it since then so I'm going to close this issue if no one else experienced the same problem it might mean that my understanding on the problem was not comprehensive.

PR #762, opened 2016-11-04
Check tokens size before deleting (does not fix issue, should be rejected)

horn-toad:
Tokens was being converted to a String when it was empty and throwing error messages.


horn-toad:
Issue was caused by a bad migration that set tokens to be a string instead of json. Checking the size of the string would not have worked properly anyway, because the empty hash was being saved in the string meaning the check would not make a difference.

Issue #749, opened 2016-10-14
OAuth (GitHub) redirects to callback url twice


neroniaky:
Hey,

I'm currently trying to the OAuth to work with Angular2-Token. You can give the OAuth a try here: [Heroku Demo](https://angular2-token.herokuapp.com/). (Angular2 + Devise Token Auth + Rails 5.0 API).

It works as expected until the last redirect:
1. Open second GitHub OAuth window for login ✅ 

`https://github.com/login/oauth/authorize?client_id=5db8f6c05510a8f5ab5f&redirect_uri=https%3A%2F%2Fangular2-token.herokuapp.com%2Fomniauth%2Fgithub%2Fcallback%3Fresource_class%3DUser&response_type=code&scope=email%2Cprofile&state=51f51db231dfea77419cb0772a97cd3d2f6d0fa9302af969`
1. After successful sign in Github redirects to `/oauth/github/callback` ✅ 

`https://angular2-token.herokuapp.com/omniauth/github/callback?code=b51d004c4a84daf0684c&resource_class=User&state=51f51db231dfea77419cb0772a97cd3d2f6d0fa9302af969`
1. Then redirect to `//oauth/github/callback` (notice the double slash) which does not exist. ❌ 

Any suggestions how to solve this? Thanks a lot!

piotrkaczmarek:
Hey @neroniaky,
I have run into the same issue when I was trying to integrate this gem with Angular2-Token too. Here I've submitted a fix for that double slash issue https://github.com/lynndylanhurley/devise_token_auth/pull/765 . Although it might not fix your problem entirely, it unblocks this one step. :)
Looking forward to hearing your feedback!

neroniaky:
Fixed with PR #765

PR #765, opened 2016-11-08
Fixes constructing redirect_route

piotrkaczmarek:
Fixes double slash in redirection url issue reported in https://github.com/lynndylanhurley/devise_token_auth/issues/749 .

The PR solves the issue but it's not critical at all - turns out redirection works fine even with the double slash. You can see it on [my example app](http://ng2auth-router-starter.netlify.com/).

Anyway, it would be nice to merge this to prevent potential future issues.

### DIFF
--- a/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb
+++ b/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb
@@ -66,6 +66,11 @@ class OmniauthTest < ActionDispatch::IntegrationTest
       get_success
     end
 
+    test 'should be redirected via valid url' do
+      get_success
+      assert_equal 'http://www.example.com/auth/facebook/callback', request.original_url
+    end
+
     describe 'with default user model' do
       before do
         get_success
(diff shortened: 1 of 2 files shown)


test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb
```diff
@@ -66,6 +66,11 @@ class OmniauthTest < ActionDispatch::IntegrationTest
       get_success
     end
 
+    test 'should be redirected via valid url' do
+      get_success
+      assert_equal 'http://www.example.com/auth/facebook/callback', request.original_url
```
piotrkaczmarek: Without changes in this PR, this spec fails as `request.original_url` equals `'http://www.example.com//auth/facebook/callback'`.

Rush:
Can we have it merged? I encountered the same issue

Rush:
Also, I confirm it fixes the problem. Will be running on my fork of devise_token_auth before this and my other PR is merged.

booleanbetrayal:
I like clean URLs too, @piotrkaczmarek . Thanks!  =]

PR #793, opened 2016-12-17
Support authenticate with facebook access token

Charlie-Hua:
Added a new condition in `OmniAuthCallbacks#render_data_or_redirect` to render user data if the OmniAuth strategy name matches 'AccessToken'.

It's supposed to work with lynndylanhurley/ng-token-auth#341 and the omniauth-facebook-access-token gem.


israelb:
Hi, @Charlie-Hua, do you have an example about how to use this with a native app? I'm trying to implement this in a project

Charlie-Hua:
@israelb I don't know how to use with native app, but this PR should work if you put `access_token=....` in the request params to the route mapping to `devise_token_auth/omniauth_callbacks#omniauth_success`
You can use `rake routes` to check which URI it is, for my project it is `/auth/:provider/callback` and thus `/auth/facebook/callback`

lynndylanhurley:
@Charlie-Hua - thanks for taking the time to send this PR. Sorry for the delay, I'll take a look ASAP!

app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
```diff
@@ -209,6 +209,10 @@ def render_data_or_redirect(message, data, user_data = {})
 
         # build and redirect to destination url
         redirect_to DeviseTokenAuth::Url.generate(auth_origin_url, data.merge(blank: true))
+
+      elsif using_access_token_strategy?
+        render json: { success: true, data: user_data.merge(data).merge({message: message}) }
```
thibaultdalban: Little suggestion : Extract this into a dedicated method, so it will be really easy to customize the response format.
Have a look at https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/registrations_controller.rb#L65 for an example.

zachfeldman:
@lynndylanhurley were you possible close on this one?

lynndylanhurley:
@zachfeldman I'm working to add support to a client to verify that this works as expected. If anyone else has the time to set up a client with usage instructions I'll be happy to check it out.

zachfeldman:
Cool thanks so much @lynndylanhurley !!

nantestdevoloper:
I changed /auth/:provider/callback url to api/v1/auth/:provider/callback and set routes for same in routes but getting auth_hash nil

zachfeldman:
@nantestdevoloper is your comment related to this pull request?

PR #800, opened 2017-01-03
Fix language errors in German locale

morgler:
Correcting a few spelling and grammatical errors as well as bad formulations in German locale.


booleanbetrayal:
thanks @morgler !

PR #826, opened 2017-02-01
Fix devise confirmation

dilnavazk:


olleolleolle:
Would you care to expand on this change? It looks weird and the tests fail.

ratschance:
I can see three immediate issues with this PR

1. The tests are failing
2. There is no description of what you are "fixing"
3. 30 lines of code are commented out, but left in the file

rmcsharry:
Looking at the changes this is an attempt to disable confirmable in this gem so that a Rails app can use normal devise gem and use :confirmable via that gem. An odd solution to the problem of having a Rails app that is both an API using tokens and a standard Rails app. I'd reject this PR.

lynndylanhurley:
@dilnavazk thanks for taking the time to submit this PR. I do agree with the above comments. Please provide an explanation for this PR so we can discuss. Closing for now.

PR #840, opened 2017-02-28
Improve documentation for testing.

JonRowe:
Hi there, thanks for your hard work on this gem, it'd be great if the documentation was a little bit more specific about how this worked, it was non obvious to me that there were 4 headers required to make this work (I was initially using the bearer token template in an authorization head (doh)).

Theres also no documentation in the readme about how to test this, you have to rely on searching through issues to find a solution. I've made a start at improving this here, but would welcome follow up suggestions.

I'm actually setting the headers individually as I'm using rack-test directly rather than test-unit style rails helpers.


lynndylanhurley:
Thanks @JonRowe!!

PR #842, opened 2017-03-01
Add Albanian locale

Phatossi:
Locale has now been translated into Albanian language as well. The new locale file with Albanian translations has been named as 'sq.yml', because 'sq' are representative letters for 'shqip' which means Albanian. These two letters are commonly widespread as representatives of Albanian language in other translations of foreign texts in Albanian (e.g. https://sq.wikipedia.org/).

### DIFF
--- a/config/locales/sq.yml
+++ b/config/locales/sq.yml
@@ -1,4 +1,4 @@
-en:
+sq:
   devise_token_auth:
     sessions:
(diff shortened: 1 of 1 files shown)


korabh:
+1 👍

booleanbetrayal:
Thanks @Phatossi !

Phatossi:
It is my pleasure @booleanbetrayal

PR #849, opened 2017-03-17
Ability to use different default fields in model

blddmnd:
Now you do not need to use required nickname, image and name fields


zachfeldman:
Interesting solution and quick to review!

lynndylanhurley:
Smart! Thx @blddmnd 👍

PR #863, opened 2017-04-07
Persist allow_password_change in the database

### LINKED ISSUE #481: @resource.allow_password_change is not persisted across requests
My application is set to require current_password when users want to change their password.
The problem is when an user wants to reset the password. After the mail is sent and the user is prompted to insert the new password, the server returns 422 with the error: "current_password can't be blank".

I'm using v0.1.37.beta4, and this version contains a pull request: https://github.com/lynndylanhurley/devise_token_auth/commit/d33bcdb5bae993bd4ba8134b52486c8b6cce2908
which is supposed to fix this problem, but it doesn't.

I guess the problem here is that @resource.allow_password change is set to true in projects_controller#edit but it's not persisted to the DB, and in the next call (projcets_controller#update) its value is false again.

I made a quick and dirty fix to store allow_password to the DB, though I'm not sure it's a good solution because this means changing the generators too....
I don't have the time to implement tests, so I didn't even open the PR. However, you can see the changes here https://github.com/Chosko/devise_token_auth/commit/e3a4d3d888547662352c8bc4568ddab6b84ac259.

MohamedBassem:
Fixes #481 
`@resource.allow_password_change` is not persisted across requests which makes it impossible to reset a user's password after a password reset when `config.check_current_password_before_update` is set to `:password`.
This pull request:
1- Modifies the generators to add the new column.
2- Adds a new test to cover this case.
3- Removes the old attribute accessors.

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -125,6 +125,7 @@ module DeviseTokenAuth
 
       if @resource.send(resource_update_method, password_resource_params)
         @resource.allow_password_change = false
+        @resource.save!
 
         yield @resource if block_given?
         return render_update_success
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -41,12 +41,6 @@ module DeviseTokenAuth::Concerns::User
     # remove old tokens if password has changed
     before_save :remove_tokens_after_password_reset
 
-    # allows user to change password without current_password
-    attr_writer :allow_password_change
-    def allow_password_change
-      @allow_password_change || false
-    end
-
     # don't use default devise email validation
     def email_required?
       false
--- a/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
+++ b/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
@@ -11,6 +11,7 @@ class DeviseTokenAuthCreate<%= user_class.pluralize %> < ActiveRecord::Migration
       ## Recoverable
       t.string   :reset_password_token
       t.datetime :reset_password_sent_at
+      t.boolean  :allow_password_change, :default => false
 
       ## Rememberable
       t.datetime :remember_created_at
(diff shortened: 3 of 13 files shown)


Bernwoop:
I think you should add the new attribute to the except list at token_validation_response (User concern) so it doesn't get serialized.

Bernica:
@MohamedBassem @lynndylanhurley Any news on this?

zachfeldman:
Makes sense to me and no merge conflicts. Didn't review the code but will try to come back to this soon has anyone else reviewed the code here? @Bernica perhaps you can commit the change you recommended?

lynndylanhurley:
I'm cleaning up the test suite and the [original test](https://github.com/lynndylanhurley/devise_token_auth/blob/62db57d58d2d61e325d3b4b41fe37b6f027fae14/test/controllers/devise_token_auth/passwords_controller_test.rb#L319) around this feature is now failing. Looking at this now and I'm not sure how it was ever passing.

misham:
I know this has already been merged in and I'm coming late to the party but how are existing users supposed to leverage this now?  Is the expectation to regenerate the User migration or .... ?

Some kind of docs on this would be amazing.  Happy to write something up, but given the current state of the code, the only way forward I see is a manually generated migration.

If this is the expected path forward, let me know and I can create a wiki page.

lynndylanhurley:
@misham - that's a good question. should we include a new migration for the fix? how is this situation usually handled by other gems?

misham:
The "easy button" option would be a separate migration that is created for existing installs for the default case.  That is, User model exists with the default columns.

Bonus points would be to use the configuration in the existing app to detect what the model is/are and create migration(s) for everything that's needed.

If the above is not feasible, then an upgrade section in the README would be the next best thing.

One caveat to the above, I solved this by implementing a version of the solution proposed here: https://github.com/lynndylanhurley/devise_token_auth/issues/481#issuecomment-244594958

So I had to back my changes out before I anything else.  Not sure how that plays into this though.

zachfeldman:
@misham I'd be down to see a pull request for either that migration and/or an upgrade to the README. Thanks for your suggestions.

PR #864, opened 2017-04-08
use URI::HTTPS to generate HTTPS redirects

cgc:
When deploying my app, I noticed the callback URL (after this redirect) was stalling, trying to load `http://my-app.herokuapp.com:443/api/auth/twitter/callback`. Playing with `URI::HTTP.build` and `URI::HTTPS.build`, I noticed the `scheme` kwarg is ignored
```ruby
> URI::HTTP.build(scheme: 'http', host: 'hi.com', port: 80).to_s
=> "http://hi.com"
> URI::HTTP.build(scheme: 'https', host: 'hi.com', port: 443).to_s
=> "http://hi.com:443"
> URI::HTTPS.build(scheme: 'http', host: 'hi.com', port: 80).to_s
=> "https://hi.com:80"
> URI::HTTPS.build(scheme: 'https', host: 'hi.com', port: 443).to_s
=> "https://hi.com"
```

It seems like the best solution is to switch which `build()` we are using based on the request scheme. I modeled my change after this PR https://github.com/fog/fog/pull/2159 which seems to be the same issue. I was unable to write a test because I don't think `request_via_redirect` or `get_via_redirect` supports having a scheme like `https`.


booleanbetrayal:
LGTM! Will look at coverage after it lands. Thanks @cgc !

PR #865, opened 2017-04-11
Use rails validator instead of custom one

### LINKED ISSUE #833: Incompatibility with shoulda in email uniqueness
If I add `include DeviseTokenAuth::Concerns::User` to user, shoulda complains with `User validations should validate that :email is case-insensitively unique`. Weird because the email is validated as unique if I check with a manual test instead of shoulda.

MaicolBen:
Resolves #833

Shouldn't be `uid` validated unique with scope in provider? And the email case insensitive ?


lynndylanhurley:
Thanks @MaicolBen!

PR #872, opened 2017-04-15
Updated generator test code to work with rails 5

jrhee17:
Prior to adding generator support for mongoid, I fixed some broken tests when running with rails 5.

Tested also with rails 4.2 (found that <=4.1 isn't supported due to dependencies)

Also got rid of some deprecation warnings and changed some indenting -- not sure about conventions for this project yet=(

Let me know if anything should be changed


booleanbetrayal:
thanks @jrhee17 !

PR #903, opened 2017-06-10
DeviseTokenAuth.bypass_sign_in setting

Borzik:
This PR adds a configuration to force passing sign in hooks on each call which includes access token. 

This may be useful, for example, for `activatable` [hook implementation](https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/activatable.rb) (you add `active` column to users table, and set it to `true` when you want to allow people to log in, but set this to `false` when you want to block them).

Default behavior remains the same.

When `DeviseTokenAuth.bypass_sign_in` is set to `false`, users which return `false` to `active_for_authentication?` will be locked immediately, and won't be able to run any actions using their access token, even if it was valid before their inactivation.

Please let me know if you have any questions or comments regarding this implementation.


Borzik:
It looks like although this implementation seems to be working, it requires additional changes, because right now it increments sign_in_count and updates last_sign_in_at and other fields on each request, which is not acceptable

PR #911, opened 2017-06-21
bypass_sign_in setting (re-created)

Borzik:
This PR adds a configuration to force passing warden fetch hooks on each call which includes access token. 

This may be useful, for example, for `activatable` [hook implementation](https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/activatable.rb) (you add `active` column to users table, and set it to `true` when you want to allow people to log in, but set this to `false` when you want to block them).

Default behavior remains the same.

When `DeviseTokenAuth.bypass_sign_in` is set to `false`, users which return `false` to `active_for_authentication?` will be locked immediately, and won't be able to run any actions using their access token, even if it was valid before their inactivation.

We use `event: :fetch`, because we want to make warden "think" we are getting user from session (this is the only way to make it skip `trackable`, `lockable`, and other sign-in-related hooks).

Please let me know if you have any questions or comments regarding this implementation.


zachfeldman:
This seems like useful behavior to me with decent tests, with all tests passing. Anybody have objections to merging this? Thanks for the PR @Borzik .

PR #914, opened 2017-06-23
Feature/allow omini json override

thatandyrose:
Hi there!

While using `devise_token_auth` with `Omniauth` I hit onto an issue where the user json being returned didn't match the schema used on my frontend vuejs app.

I couldn't find an easy way to "hook" into the gem without forking and/or copying and pasting the`omniauth_success` contents, which I didn't want to do because of possible future breaking changes.

I know there is a hook in there to pass a block, but if I somehow did that in there (like providing a proxy object or something)... it just feels wrong since it assumes knowledge of the underlying gem implementation.

Anyhow, this provide a simple method that can be overriden which is just for providing some way to serialize the resource object.

Yay 👍  or nay 👎 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,6 +729,24 @@ mount_devise_token_auth_for 'User', at: 'auth', controllers: {
 }
 ~~~
 
+## Overriding the Omniauth Success resource serializer
+
+Rendering an Omniauth success works a litte differently to the other controller which is why there is no render override. Instead we provide a method to override which defines how you want the resource (usually the user object) to be serialized. The default is:
+
+~~~ruby
+def omniauth_resource_serializer(resource)
+  resource.as_json
+end
+~~~
+
+But you may not want to use the built in `as_json`. In which case, you can override it like this, for example:
+
+~~~ruby
+def omniauth_resource_serializer(resource)
+  MyCustomSerializer.new(resource).return_a_hash
+end
+~~~
+
 **Note:** Controller overrides must implement the expected actions of the controllers that they replace.
 
 ## Passing blocks to Controllers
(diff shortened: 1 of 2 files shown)


thatandyrose:
@lynndylanhurley any thoughts on this...? If it's not good we could close it, but would be great know why?

Thanks! 👍

lynndylanhurley:
Hey @thatandyrose, sorry for the delay.

There's a method in the `User` concern called `token_validation_response` that we're using everywhere else for this:
https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L231

For example, in my apps I've been doing something like this:

```ruby
    # in app/models/user.rb

    def token_validation_response
      ActiveModelSerializers::SerializableResource.new(
        self,
        serializer: UserSerializer,
        key_transform: :camel_lower
      ).as_json
    end
```

And that way the user model is serialized the same for sign_up, registration, and token validation.

It looks like we should be using the same method for the omniauth response but we're not.

Would it be possible to just change `@resource.as_json` to `@resource.token_validation_response` on this line?
https://github.com/lynndylanhurley/devise_token_auth/blob/7f61b60bb8c524db390b00169b10970d0af9a27a/app/models/devise_token_auth/concerns/user.rb#L231

Thanks for taking the time to submit this : )

zachfeldman:
Hey @thatandyrose we'd love for you to finish this up if you have a moment! :) thanks again for your time.

thatandyrose:
hey @lynndylanhurley and @zachfeldman sorry for the delay guys!

Cool so shall I just make sure the ominauth controller is using that user concern method and then update the docs and we're good?

Thanks for getting back and again sorry for the delay!

cheers guys and gals!

lynndylanhurley:
> Cool so shall I just make sure the ominauth controller is using that user concern method and then update the docs and we're good?

That would be amazing. Thanks 👍

PR #915, opened 2017-06-27
fix batch requests update_auth_header

nathanscott:
Fix to update request headers during batch requests


nathanscott:
Scratch that, the server seems to be doing the right thing.
The browser seems to be caching response headers from previous requests.

Please disregard PR :)

PR #949, opened 2017-09-01
WIP: Multiple login fields

danielneis:
Hello,

this is a first version of a change to allow login using all authentication_keys defined in the model.

Note that this versions removes the valid_params? for simplicity and also the addition of the BINARY keyword on mysql, but if it is really needed we can work to add it back.

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -10,26 +10,28 @@ module DeviseTokenAuth
 
     def create
       # Check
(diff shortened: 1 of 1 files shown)


zachfeldman:
@danielneis looks like tests fail on this pr? Thanks for your contribution but mind taking a look at that?

danielneis:
Hello, I rebased the patch to current master and added back the "binary" statement for mysql

MaicolBen:
Closed since the other PR will take over this

PR #957, opened 2017-09-14
change devise method to reset password by token

dks17:
When I clicks the link in the email, which brings me to password reset (get /password/edit)
every time I catch error ( password: 'can't be blank') in @resource instance.

It appears from [reset_password_by_token](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L143) when [reset_password](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L41) is called.

In this case I guess using [with_reset_password_token](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L123) method lighter (checks only token and doesn't pass the error).


zachfeldman:
Seems sane. Marked as easy to merge for hopefully faster review.

MaicolBen:
Yeah, you're right, I just printed the errors of the model in the tests and it includes the password error.

PR #964, opened 2017-09-27
Allow other provider than email when logins

### LINKED ISSUE #859: How to authenticate user using username using this gem ?

### LINKED ISSUE #893: Change provider when sessions controller
In ```devise_token_auth/app/controllers/devise_token_auth/sessions_controller.rb```
I see ```q = "#{field.to_s} = ? AND provider='email'"```

But I don't use provider is email. Could you change it.

MaicolBen:
Resolves #893
Resolves #859

I am not sure about adding a method overridable in the controller, maybe it should a constant in user.

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -20,7 +20,7 @@ module DeviseTokenAuth
           q_value.downcase!
         end
 
-        q = "#{field.to_s} = ? AND provider='email'"
+        q = "#{field.to_s} = ? AND provider='#{provider}'"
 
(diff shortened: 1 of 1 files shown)


lynndylanhurley:
Thanks @MaicolBen!!

codeundercoverdev:
Where actually is the answer to this question?  I have been getting redirected from github issue to github issue saying the answer is somewhere but all I see is a list of github issues refering the problem with zero code written anywhere?

SanketKarve:
You need to override the provider method in session controller. This works for me: 
```
def provider
  super
  'custom-provider-name'
end

```

PR #971, opened 2017-10-09
Adding in unlocks controller and specs. This should resolve #927.

### LINKED ISSUE #927: Can't unlock account through email link in lockable
I was implementing lockable but it threw 500 error if I click on the email link. 
I rake route and I found that it goes on **Devise::UnlocksController** and I checked for **UnlocksController** in **devise_token_auth** and I didn't find one. 
Do I have to override unlocks controller or am I missing something else?

brycesenz:
This should resolve the core issue of #927.  However, it has these known shortcomings (and perhaps more that I haven't identified):

1. The redirect after an unlock is just the root path (i.e. '/'), and it is set in a controller method.  I'm not as familiar with how the library does configuration and default handling, but this would make more sense as a configurable option.
2. The messages for the Unlocks controller only have English translations in the locales

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -1,6 +1,7 @@
 module DeviseTokenAuth
   class ApplicationController < DeviseController
     include DeviseTokenAuth::Concerns::SetUserByToken
+    include DeviseTokenAuth::Concerns::ResourceFinder
 
     def resource_data(opts={})
       response_data = opts[:resource_json] || @resource.as_json
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -1,0 +1,35 @@
+module DeviseTokenAuth::Concerns::ResourceFinder
+  extend ActiveSupport::Concern
+  include DeviseTokenAuth::Controllers::Helpers
+
+  def get_case_insensitive_field_from_resource_params(field)
+    # honor Devise configuration for case_insensitive keys
+    q_value = resource_params[field.to_sym]
+
+    if resource_class.case_insensitive_keys.include?(field.to_sym)
+      q_value.downcase!
+    end
+    q_value
+  end
+
+  def find_resource(field, value, provider='email')
+    # fix for mysql default case insensitivity
+    q = "#{field.to_s} = ? AND provider='#{provider.to_s}'"
+
+    if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+      q = "BINARY " + q
+    end
+
+    @resource = resource_class.where(q, value).first
+  end
+
+  def resource_class(m=nil)
+    if m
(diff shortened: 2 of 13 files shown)


brycesenz:
@zachfeldman - oh, haha, probably just human error.  Like I said, I'm awful with Minitest - I couldn't figure out the command to test just one spec in isolation (as opposed to the whole file), so I took to commenting things out once they worked.  Please let me know the rest of your thoughts once you've had a chance to review.

Also, as should be clear from the code, I added an API endpoint to let a user trigger an unlock email to be sent to his/her account (a `create` method in the controller).  The specs cover that as well, but I do want to call it out since it's functionality beyond the scope just of this issue.

zachfeldman:
No problem, mind uncommenting that so we can run Travis against it?

Yeah that sounds like a code thing to have! Thanks for pointing it out.

zachfeldman:
All checks have passed. Going to approve this - if one more person does, I will merge.

MaicolBen:
Nice work there, I feel you about Minitest, however, you did great testing! 

But I don't like some copy/pasted code from `SessionsController`. In fact you didn't take into account a recent change in the providers' query: #964

Can you in some way refactor the first part of the create method to a concern or to the `DeviseTokenAuth::ApplicationController` ? Let us know if you can't.

brycesenz:
@MaicolBen - Haha, yeah, I ripped a lot of stuff from the PasswordsController actually.  The two are very similar!  That provider change has actually not made its way into the PasswordsController yet I see - good reason to abstract it!  Is that the only part of the create action that you were talking about refactoring?  It seems like a fair bit of the library could use some refactoring (no offense; that's the nature of open source sometimes), but I don't want to have the scope of this PR bleed too far.

brycesenz:
Ok, submitted a refactor with a new module.  Some parts of the code still aren't really clear to me - for example, this block appears in the`ApplicationController`, but if I try to move it to a separate module, specs start failing.  And minitest unfortunately gives me so many warnings that it's hard to track down what's actually going wrong.

```ruby
    def resource_class(m=nil)
      if m
        mapping = Devise.mappings[m]
      else
        mapping = Devise.mappings[resource_name] || Devise.mappings.values.first
      end

      mapping.to
    end
```

MaicolBen:
You didn't give me time to tell you to not do it if it were too much effort! But you did great work! I will review it. I'm thinking in cleaning it more in the future.

lynndylanhurley:
>It seems like a fair bit of the library could use some refactoring

Agreed. The code climate score for this project is really embarrassing x_x

brycesenz:
@lynndylanhurley - But the test coverage is solid!  I mean for real - I would not have had the confidence to change anything if I didn't have a test suite to run against to make sure I didn't break anything.  Y'all have done an awesome job keeping the test suite in order!

brycesenz:
@MaicolBen - Both of those points are valid.  I have incorporated those changes into the PR.

zachfeldman:
Looks good to me! Love the `ResourceFinder` abstraction. Merging

PR #975, opened 2017-10-11
Fix default provider after refactor in concern

MaicolBen:
This PR #971 broke my what I did on #964, because it added the provider as a default parameter instead of a overridable method. I can't add test with other providers because the method needs to be override (something done in a project that includes this gem), and I don't like to do monkeypatch testing.


zachfeldman:
Makes sense to me

Issue #959, opened 2017-09-15
Why does a confirm_success_url column appear in some of the test migrations?


ethagnawl:
From [db/migrate/20140715061447_devise_token_auth_create_users.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/536911c7d3f178637423b77fe1179388ddf62b9e/test/dummy/db/migrate/20140715061447_devise_token_auth_create_users.rb#L25-L30):

```
      ## Confirmable
      t.string   :confirmation_token
      t.datetime :confirmed_at
      t.datetime :confirmation_sent_at
      t.string   :confirm_success_url
      t.string :unconfirmed_email # Only if using reconfirmable
```

I saw this while grepping the codebase and - naively - assumed that if I added that column to my user table, the gem would reference that value when sending emails. Unsurprisingly, this is not the case.

I'd propose updating those migrations if those columns are based on features which no longer exist or, if they're there for good reason, adding documentation to explain their purpose.

zachfeldman:
@ethagnawl I think that's probably a wise idea? Perhaps a PR is in order!

MaicolBen:
I don't know if you can check the user's modules in order to add specific fields in the migration, maybe you can document each module's field [like devise](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/generators/active_record/devise_generator.rb)

ethagnawl:
Closed via #978.

PR #978, opened 2017-10-12
remove confirm_success_url entries from dummy migrations to prevent c…

ethagnawl:
…onfusion 

See #959.


zachfeldman:
I see no problem with this. Waiting on one other to confirm, then I'll merge!

PR #983, opened 2017-10-18
Fix user tokens that don't properly deserialize

### LINKED ISSUE #121: User tokens don't properly deserialize

MaicolBen:
Resolves #121

### DIFF
--- a/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
+++ b/lib/generators/devise_token_auth/templates/devise_token_auth_create_users.rb.erb
@@ -43,6 +43,16 @@ class DeviseTokenAuthCreate<%= user_class.pluralize %> < ActiveRecord::Migration
       ## Tokens
       <%= json_supported_database? ? 't.json :tokens' : 't.text :tokens' %>
 
+      reversible do |direction|
+        direction.up do
+          <%= user_class %>.find_each do |auth|
+            auth.uid = auth.email
+            auth.tokens = nil
+            auth.save!
+          end
+        end
+      end
+
       t.timestamps
     end
 


zachfeldman:
@MaicolBen I don't exactly understand the error and why this fixes it - if you can explain I'm glad to review and possibly merge!

MaicolBen:
@zachfeldman I don't know the cause of the error. The token is stored corrupted randomly, so once it tries to read `tokens` in the [login](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/sessions_controller.rb#L32), it breaks with `string not matched`. As an example of a wrong stored token:
```
"{\"70crCAAYmNP1xLkKKM09zA\":{\"token\":\"$2a$10$mSePnpVMaaegCpn3AhORGe5wajFhgMoBjGIrMwq4Qq2mP6f/OHu1y\",\"expiry\":12121234}}"
```

I know this isn't the best solution not knowing the cause, and since it's almost impossible to reproduce it, this is the solution that solved the problem. You can close this PR but I think the issue should remain open.

zachfeldman:
Hmm ok I'll defer to @lynndylanhurley or another contributor on this one then.

lynndylanhurley:
I'll take a look ASAP. I'm not sure if it's possible to write tests around this, but I'll try to trigger the issue and maybe see what's causing it.

MaicolBen:
Closing this since now I realized that this migration won't make sense without users (it just created the table).

PR #985, opened 2017-10-20
Add link to wiki of how to add fields for an existing user table

### LINKED ISSUE #181: Best way to set up migration for installation on existing User table already using Devise?
I am attempting to install `devise_token_auth` but almost everything I've read thus far seems to refer to a freshly created Rails app.

Here's the migration I came up with, but it seems like I'm leaving out the various validations and indexes that exist in the `devise_token_auth` CreateUser migration.

``` ruby
class DeviseTokenAuthChangeUsers < ActiveRecord::Migration
  def change
    change_table :users do |t|
      t.string :provider
      t.string :uid
      t.text :tokens
    end
  end
end
```

Here's an example of one of the errors I received when trying to migrate with an index on a field that's obviously going to be blank with existing users. I'm not sure what I'm supposed to do in the migration to ensure uid is unique for existing users is part of the problem.

``` ruby
PG::UniqueViolation: ERROR:  could not create unique index "index_users_on_uid"
DETAIL:  Key (uid)=() is duplicated.
```

MaicolBen:
Resolves #181 

Thanks to @rruettimann !

You can make here any comments about the [wiki](https://github.com/lynndylanhurley/devise_token_auth/wiki/Setup-migrations-for-an-existing-User-table), and I will address them. If you are a contributor, you can edit it.

Note: we should move more things from the README to the wiki, it's too long

### DIFF
--- a/README.md
+++ b/README.md
@@ -847,6 +847,11 @@ end
 
 You may be interested in [GrapeTokenAuth](https://github.com/mcordell/grape_token_auth) or [GrapeDeviseTokenAuth](https://github.com/mcordell/grape_devise_token_auth).
 
+### I already have an user, how can I add the new fields?
+
+Check [Setup migrations for an existing User table](https://github.com/lynndylanhurley/devise_token_auth/wiki/Setup-migrations-for-an-existing-User-table)
+
+
 # Conceptual
 
 None of the following information is required to use this gem, but read on if you're curious.


zachfeldman:
Nice! I agree we need to add more stuff to the Wiki from the README. Merging as it doesn't affect the core codebase.

karlingen:
Inside: 

```ruby
User.find_each do |user|
  user.uid = user.email
  user.provider = 'email'
  user.save!
end
```
May I ask why you've chosen this approach: `user.uid = user.email` instead of `user.uid = user.id`?

MaicolBen:
uid is the identifier of the provider, in case of email is the email, in case of facebook, is the facebook id

PR #989, opened 2017-10-21
Refactor render error

### LINKED ISSUE #987: Password Reset: Is it possible to disable returning 404 when user is not found?
I'm trying not having to re-implement the entire controller functionality and just want to change the behavior a bit.

I don't care if the user is found or not, the response should always be 200.

Is the only solution to just re-implement `POST create` for the passwords controller?

Thank you

MaicolBen:
This PR includes:
- Resolves #987, now it's possible edit/disable 404 error in passwords#create
- Refactor the errors rendering (not the resource's errors), so it's possible to change how to return them. In my opinion, we shouldn't be using `errors` for an array with a message, in the frontend you have to ask if it's an array vs hash (instead of asking error vs errors), it's kind of ugly. As I don't know if I can make that change (because I can break ng-token-auth), I moved how the errors are returned to a method.
- Removed @error_status and @errors from passwords/unlocks controller, they made the possible flows more complex
- Added Ruby 2.2 and 2.4 to Travis, so we can support more rubys.
- Renamed `is_json_api`, following the [ruby style guide](https://github.com/bbatsov/ruby-style-guide#bool-methods-qmark)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ cache: bundler
 sudo: false
 
 rvm:
+  - 2.2.8
   - 2.3.1
+  - 2.4.2
 
 env:
   global:
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -5,7 +5,7 @@ module DeviseTokenAuth
 
     def resource_data(opts={})
       response_data = opts[:resource_json] || @resource.as_json
-      if is_json_api
+      if json_api?
         response_data['type'] = @resource.class.name.parameterize
       end
       response_data
@@ -48,7 +48,7 @@ module DeviseTokenAuth
       mapping.to
     end
 
-    def is_json_api
+    def json_api?
       return false unless defined?(ActiveModel::Serializer)
       return ActiveModel::Serializer.setup do |config|
         config.adapter == :json_api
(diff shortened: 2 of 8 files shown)


zachfeldman:
I don't see any problems with this PR, looks good to me. Will wait for one more positive review before merging though!

zachfeldman:
@MaicolBen I can merge this if conflicts are solved!

lynndylanhurley:
These changes look good and are definitely an improvement. My only concern is that changing the error format will break compatibility with existing clients.

@MaicolBen - what do you think about adding a deprecation warning to the old format for the next release, and then switching over to a new format by default in the following release?

MaicolBen:
@zachfeldman Rebased

@lynndylanhurley I don't know if the deprecation could help here because the frontend won't change it until we change how we send the responses (it isn't like the methods or fields removal).

zachfeldman:
Thanks @MaicolBen ! Looks good to merge to me but will wait to merge until @lynndylanhurley blesses it or decides we need a deprecation warning.

lynndylanhurley:
>will wait to merge until @lynndylanhurley blesses it or decides we need a deprecation warning.

Has the error format changed with this PR? Maybe I misunderstood but I thought this PR just provided an easier way to override the error format.

MaicolBen:
@lynndylanhurley As you said, this PR doesn't change the error format (you can see there aren't test changes), only provides an easy way. In the description, I recommend not using arrays for an error message, but that can be done in another PR for the next release.

lynndylanhurley:
Cool, that's what I thought, just making sure.

This looks good to merge. Let's address the error format issue in another thread.

PR #994, opened 2017-10-24
Remove save tokens in build auth header

### LINKED ISSUE #992: Improve after action update headers in case change headers is disabled
In the [after action](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L92) when the headers update is disabled, it stores the token that is retrieved by the frontend which must be the same since is checked in the before action, why is that?

The code that I'm talking about is: 
```ruby
    if @used_auth_by_token && !DeviseTokenAuth.change_headers_on_each_request
      # should not append auth header if @resource related token was
      # cleared by sign out in the meantime
      return if @resource.reload.tokens[@client_id].nil?

      auth_header = @resource.build_auth_header(@token, @client_id)

      # update the response header
      response.headers.merge!(auth_header)
```

I think we can get rid of that code and improve the performance (and thus removing 3 db requests each time an api request is made)

MaicolBen:
Resolves #992 

This will improve the performance of each request in case `change_headers_on_each_request` is disabled.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -199,24 +199,14 @@ module DeviseTokenAuth::Concerns::User
       updated_at: Time.now
     }
 
-    return build_auth_header(token, client_id)
+    return update_auth_header(token, client_id)
   end
 
 
   def build_auth_header(token, client_id='default')
-    client_id ||= 'default'
-
     # client may use expiry to prevent validation request if expired
     # must be cast as string or headers will break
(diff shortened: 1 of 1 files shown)


zachfeldman:
@MaicolBen since this hasn't moved or had any disapproval, I'd feel free to merge if you're ready!

MaicolBen:
@zachfeldman I think is good to go!

lynndylanhurley:
Looks good, thanks @MaicolBen!

Issue #993, opened 2017-10-23
Confirmation URL query parameters don't match header counterparts


ethagnawl:
I've noticed that the keys used in the query parameters added to my email confirmation mailer's `redirect_url` don't match their header counterparts. (I'm currently using v0.1.42.)

Here's an example:

`http://www.somewhere.com/?account_confirmation_success=true&client_id=xxx&config=default&expiry=yyy&token=zzz&uid=someone@somewhere.com`

I'm writing documentation which will instruct FE devs on how to submit a PATCH request to update the confirmed user's account (user is invited to application and then uses this confirmation page to enter additional account info - I'd like to have this be a single step instead of them setting a password, logging in and then entering info). It seems counter-intuitive to say, "extract `client_id|token` from the query string and use its value to populate the `client|access-token` header".

I'd be happy to provide any additional information and will look into submitting a backwards compatible PR if I can find some time.

zachfeldman:
@ethagnawl that's a good point. I agree that the keys should match everywhere.

I'd be very open to a PR fixing this! Thanks for reporting.

ethagnawl:
@zachfeldman I've made a first pass at this. Please have a look when you get a chance and let me know if you have any feedback.

ethagnawl:
Closed via #996.

Issue #976, opened 2017-10-12
Fix Warnings in Test Suite Output


ethagnawl:
While running the test suite (using Ruby 2.2.5 and 2.4.0) there are hundreds/thousands of lines worth of warnings output.

Some examples:

> devise_token_auth/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb:21:in `get_parsed_data_json': warning: URI.une
scape is obsolete                                                                                                                                             

> devise_token_auth/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:81: warning: instance variable @resource not initialize
d

> devise_token_auth/app/controllers/devise_token_auth/concerns/set_user_by_token.rb:81: warning: instance variable @resource not initialize
d

> bundler/gems/omniauth-facebook-19634473820d/lib/omniauth/strategies/facebook.rb:79: warning: instance variable @authorization
_code_from_signed_request_in_cookie not initialized

They distract from the task at hand and, where possible, should be addressed.

![screenshot 2017-10-12 11 18 40](https://user-images.githubusercontent.com/289949/31503981-20afbbce-af3f-11e7-9fb4-b706ef50796b.png)

zachfeldman:
@ethagnawl agreed, this is not helping when running tests. Would you have time to put together a PR that addresses this, perhaps? Thanks for bringing this to our attention!

ethagnawl:
@zachfeldman Sure. I can start chipping away at the infractions.

zachfeldman:
Urock!

lynndylanhurley:
I'm working on this now. Some of these warnings are legit, but it looks like a lot of them are coming from outdated test deps. I'll post a PR in a bit.

zachfeldman:
Thanks @lynndylanhurley !

MaicolBen:
Thanks for removing some of them! But still there are tons of warnings, the main ones are:
- `lib/devise/models/confirmable.rb:297: warning: instance variable @reconfirmation_required not initialized` (from devise I guess)
- `warning: method redefined;` because a lot of metaprograming, I don't know how to remove them
- `warning: loading in progress, circular require considered harmful` because including `require File.expand_path('../dummy/config/environment', __FILE__)`, also, I don't know how to remove them

You can omit them with `RUBYOPT=W0 rake test`

gitcoinbot:
__This issue now has a funding of 0.08 ETH (61.36 USD) attached to it.__

[...]

PR #996, opened 2017-10-24
993 - mirror auth header keys in build_auth_url query params

ethagnawl:
Addresses #993.

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -20,19 +20,12 @@ module DeviseTokenAuth
 
         yield @resource if block_given?
 
-        redirect_to(@resource.build_auth_url(params[:redirect_url], {
-          DeviseTokenAuth.headers_names["access-token"] => token,
-          DeviseTokenAuth.headers_names["client"] => client_id,
-
-          :account_confirmation_success => true,
-          :config => params[:config],
-
-          # Legacy parameters which may be removed in a future release.
-          # Consider using "client" and "access-token" in client code.
-          # See: github.com/lynndylanhurley/devise_token_auth/issues/993
-          :client_id => client_id,
-          :token => token
-        }))
+        redirect_header_options = {account_confirmation_success: true}
+        redirect_headers = build_redirect_headers(token,
+                                                  client_id,
+                                                  redirect_header_options)
+        redirect_to(@resource.build_auth_url(params[:redirect_url],
+                                             redirect_headers))
       else
         raise ActionController::RoutingError.new('Not Found')
       end
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -21,10 +21,12 @@ module DeviseTokenAuth
         yield @resource if block_given?
 
         redirect_to(@resource.build_auth_url(params[:redirect_url], {
-          token:                        token,
-          client_id:                    client_id,
+          "access-token":               token,
           account_confirmation_success: true,
-          config:                       params[:config]
+          client:                       client_id,
+          client_id:                    client_id,
+          config:                       params[:config],
+          token:                        token
         }))
       else
         raise ActionController::RoutingError.new('Not Found')
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -21,12 +21,17 @@ module DeviseTokenAuth
         yield @resource if block_given?
 
         redirect_to(@resource.build_auth_url(params[:redirect_url], {
-          "access-token":               token,
-          account_confirmation_success: true,
-          client:                       client_id,
-          client_id:                    client_id,
-          config:                       params[:config],
-          token:                        token
+          DeviseTokenAuth.headers_names["access-token"] => token,
+          DeviseTokenAuth.headers_names["client"] => client_id,
+
+          :account_confirmation_success => true,
+          :config => params[:config],
+
+          # Legacy parameters which may be removed in a future release.
+          # Consider using "client" and "access-token" in client code.
+          # See: github.com/lynndylanhurley/devise_token_auth/issues/993
+          :client_id => client_id,
+          :token => token
         }))
       else
         raise ActionController::RoutingError.new('Not Found')
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -17,6 +17,20 @@ module DeviseTokenAuth
 
     protected
 
+    def build_redirect_headers(access_token, client, redirect_header_options = {})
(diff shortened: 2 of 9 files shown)


app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -21,10 +21,12 @@ def show
         yield @resource if block_given?
 
         redirect_to(@resource.build_auth_url(params[:redirect_url], {
-          token:                        token,
-          client_id:                    client_id,
+          "access-token":               token,
```
zachfeldman: Should we maybe using the values here:

https://github.com/lynndylanhurley/devise_token_auth/blob/536911c7d3f178637423b77fe1179388ddf62b9e/lib/devise_token_auth/engine.rb#L38

Like they're used here?
https://github.com/lynndylanhurley/devise_token_auth/blob/ae87cae3a24869a5aceab5addc3f3c1041bf39d3/app/models/devise_token_auth/concerns/user.rb#L222

Just an idea.

ethagnawl: Are you referring to the hash key style? If so, they _should_ behave the same way: 

`{:"access-token" => "foo"} == {"access-token": "foo"} # true` 

Also, to use the former, we'd need to convert the rest of the hash keys to the hashrocket style.

zachfeldman: No, I guess I just mean that maybe we should use the configured keys:

`DeviseTokenAuth.headers_names[:"access-token"]`

rather than assume the user is using` access-token`. DeviseTokenAuth.headers_names seems to allow you to override the header names. I don't really know how many people are using that feature, probably not a lot because it's not used across the library.

ethagnawl: Ah, sorry - it's still early in EST. 😛 

I think that's a good idea. What would you propose we do in the case where keys aren't present in `DeviseTokenAuth.headers_names`, like `client_id`? Should those be added to `headers_names` in lib/devise_token_auth/engine.rb, even though they might be phased out at some point? (As I'd probably argue they should.)

zachfeldman: I would say things that are going to be phased out possibly in the future shouldn't get this treatment - I'd be okay with you only doing this with the essential keys, basically those already in DeviseTokenAuth.headers_names.

@MaicolBen perhaps you have an opinion on this too?

app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -21,12 +21,17 @@ def show
...
         redirect_to(@resource.build_auth_url(params[:redirect_url], {
-          "access-token":               token,
-          account_confirmation_success: true,
-          client:                       client_id,
-          client_id:                    client_id,
-          config:                       params[:config],
-          token:                        token
+          DeviseTokenAuth.headers_names["access-token"] => token,
+          DeviseTokenAuth.headers_names["client"] => client_id,
+
```
zachfeldman: Bit of a nitpick but any reason for an extra space here?

app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -21,12 +21,17 @@ def show
...
+          :account_confirmation_success => true,
+          :config => params[:config],
+
+          # Legacy parameters which may be removed in a future release.
+          # Consider using "client" and "access-token" in client code.
+          # See: github.com/lynndylanhurley/devise_token_auth/issues/993
```
zachfeldman: Beautiful

zachfeldman:
Ah sorry @ethagnawl I neglected to check the tests - looks like we have a failing test somewhere in there.

app/controllers/devise_token_auth/confirmations_controller.rb
ethagnawl: I was attempting to make a distinction between the keys coming from `headers_names` and the bare keys - I don't feel strongly about the need for that, though.

zachfeldman: Ah not a bad idea. I'm fine with this then.

MaicolBen:
Thank you for doing this, I noticed it in the past in passwords#edit. Can you in some way refactor the repeated hash of the redirects? You might add the new parameter (`unlock`, `reset_password`) to the hash in each controller.

ethagnawl:
@zachfeldman That first failure was legitimate, but I've fixed it and I'm still seeing those failures. I've been seeing intermittent failures locally (on master, in addition to my branch), so maybe that's what's going on? Could you try restarting the build?

@MaicolBen That's a great idea. Where would you suggesting putting that helper method?

zachfeldman:
Will try to restart the build now!

zachfeldman:
I don't know, I'm seeing 17 failures on the test suite...seems atypical?

MaicolBen:
You can restart it inside travis, the test runs always successfully in local computer, so it's weird. 17 failures isn't atypical, something here broke the build. 

About the refactor, I suggest adding to the application controller:
```ruby
def build_redirect_headers(token, client_id)
  {
    DeviseTokenAuth.headers_names["access-token"] => token,
    DeviseTokenAuth.headers_names["client"] => client_id,
    config: params[:config],

    # Legacy parameters which may be removed in a future release.
    # Consider using "client" and "access-token" in client code.
    # See: github.com/lynndylanhurley/devise_token_auth/issues/993
    client_id: client_id,
    token: token    
  }
end
```

In `app/controllers/devise_token_auth/confirmations_controller.rb` do `build_redirect_headers(token, client_id).merge(account_confirmation_success: true)`

In `app/controllers/devise_token_auth/passwords_controller.rb` do `build_redirect_headers(token, client_id).merge(reset_password: true)`

In `app/controllers/devise_token_auth/unlocks_controller.rb` do `build_redirect_headers(token, client_id).merge(unlock: true)`

ethagnawl:
@MaicolBen Okay, that makes sense. I will take another pass when I find a few minutes.

As I mentioned above, I'm also seeing those failures locally, on master. Interestingly, the number of failures/errors isn't consistent.

screenshot 2017-10-25 18 20 21
screenshot 2017-10-25 18 18 34

MaicolBen:
I downloaded your branch and it's failing the tests, did you test it? There are 2 mistakes:
- The headers should be symbols (`:"access-token"` and `:"client"`)
- In the passwords controller, it's `client` instead of `client_id`

ethagnawl:
@MaicolBen Thank you for pointing that out. In answer to your question: yes, I have been running the test suite. However, (per #976) it's very difficult to find errors because of the wall of output generated by the test suite.

Anyways, I've introduced that helper method we'd discussed and the test suite now seems to be passing. Please let me know if you have any feedback about the refactoring.

zachfeldman:
Test suite is passing!! Nice work @ethagnawl :)

@MaicolBen can you confirm that this PR is ready to go? If so, I will merge.

Issue #997, opened 2017-10-27
Hardcoded resource.provider in RegistrationsController?


m4-miranda:
The resource.provider field in Registrations#create is hardcoded to 'email'. Shouldn't this be obtained from Devise?

I'm trying to get rid of email dependency altogether

zachfeldman:
Hey @m4-miranda sure, I agree with your logic here. Mind opening a pull request to illustrate your exact changes in mind?

MaicolBen:
If you make a PR, you can reuse the `provider` method defined in #975

m4-miranda:
@MaicolBen It isn't possible to override the concern, is it? I could modify the get_resource method to accept a provider parameter, but the method would have to be declared in each controller

MaicolBen:
You can override it in your code, and yes, you have to do it in every controller (you can make a concern in your code to avoid repeating)

MaicolBen:
Closed via #998

KelseyDH:
This gotcha really screwed me up.  It would be really nice for `provider` not to be hardcoded from params.  e.g. by checking if its defined *before* hardcoding it to `email`

I lost many hours trying to figure out why my user model kept hardcoding provider to 'email'.

MaicolBen:
Sorry to hear that, as we don't have a lot of code documentation, you need to read the source code, in case of the provider, you can override with what was defined in #975

KelseyDH:
@MaicolBen  Thanks for getting back.  My intention is to have `provider` [switch between `username` and `email` based on the params received](https://github.com/lynndylanhurley/devise_token_auth/issues/1066) within the same controller.  Any idea on how I could go about implementing that kind of dynamic override for the registrations controller -- and not have this hardcoded override break my User model elsewhere on subsequent updates?

KelseyDH:
I tried modifying the gem to override the registrations controller code from something like this:

```
module DeviseTokenAuth
  class RegistrationsController < DeviseTokenAuth::ApplicationController
 ...
    def create
      @resource            = resource_class.new(sign_up_params.except(:confirm_success_url))
      @resource.provider   = provider
```

over to this:

```
if sign_up_params[:provider].present?
  @resource.provider = sign_up_params[:provider]
else
  @resource.provider = provider
end
```

 but it didn't work and I still get back an 'email' provider despite sending in params for `provider` and whitlisting this for `sign_up`. :(

MaicolBen:
[...]

PR #998, opened 2017-10-27
Added provider parameter to find_resource 

m4-miranda:
As discussed in comments of #997,
I've added a provider paramter to find_resource method. I've updated the calls to find_resource method in passwords controller:
```golang 
@resource = find_resource(:uid, @email, provider)
```
sessions controller:
```golang 
@resource = find_resource(field, q_value, provider)
 ```
Unlocks controller, I assume, would require an 'email' provider hence I've hardcoded the paramter there:
```golang
@resource = find_resource(:email, @email, 'email')
```
P.S I'm fairly new to devise_token_auth and devise in general so not sure about the best practices etc.
@MaicolBen @zachfeldman

### DIFF
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -12,7 +12,7 @@ module DeviseTokenAuth::Concerns::ResourceFinder
     q_value
   end
 
-  def find_resource(field, value, provider)
+  def find_resource(field, value)
     # fix for mysql default case insensitivity
     q = "#{field.to_s} = ? AND provider='#{provider.to_s}'"
     if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
@@ -30,5 +30,9 @@ module DeviseTokenAuth::Concerns::ResourceFinder
     end
 
     mapping.to
+  end
+
+  def provider
+    'email'
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/devise_token_auth/concerns/resource_finder.rb
```diff
@@ -12,10 +12,9 @@ def get_case_insensitive_field_from_resource_params(field)
     q_value
   end
 
-  def find_resource(field, value)
+  def find_resource(field, value, provider)
```
MaicolBen: I don't understand why the parameter, this won't call the method

m4-miranda: Hi, could you help me out with how to go about overriding the concern in my project that utilises this repo? I would write a concern that extends `DeviseTokenAuth::Concerns::ResourceFinder` with a ```def provider``` and override the controllers to `include customResourceFinder`. Is this right?

MaicolBen: I think that won't work, you need to create a concern with `def provider` in there and add that concern in each override controller. But if you need to add it to the registrations controller, you add it there. But you need to add PR here with only this:
screen shot 2017-10-27 at 12 56 28 pm

app/controllers/devise_token_auth/passwords_controller.rb
```diff
@@ -28,7 +28,7 @@ def create
       end
 
       @email = get_case_insensitive_field_from_resource_params(:email)
-      @resource = find_resource(:uid, @email)
+      @resource = find_resource(:uid, @email, provider)
```
MaicolBen: You forgot to remove from here

PR #1001, opened 2017-10-29
allow only one confirmation

MaicolBen:
Reopened #519 because it has conflicts and the original author can't rebase it

Thanks @rmvenancio !

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -10,6 +10,10 @@ module DeviseTokenAuth
         token_hash = BCrypt::Password.create(token)
         expiry     = (Time.now + @resource.token_lifespan).to_i
 
+        if @resource.sign_in_count > 0
+          expiry = (Time.now + 1.second).to_i
+        end
+
         @resource.tokens[client_id] = {
           token:  token_hash,
           expiry: expiry
--- a/test/controllers/devise_token_auth/confirmations_controller_test.rb
+++ b/test/controllers/devise_token_auth/confirmations_controller_test.rb
@@ -61,6 +61,12 @@ class DeviseTokenAuth::ConfirmationsControllerTest < ActionController::TestCase
         test 'User shoud have the Last checkin filled' do
           assert @resource.last_sign_in_at?
         end
+        
+        test 'user already confirmed' do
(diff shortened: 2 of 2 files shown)


app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -10,6 +10,10 @@ def show
         token_hash = BCrypt::Password.create(token)
         expiry     = (Time.now + @resource.token_lifespan).to_i
 
+        if @resource.sign_in_count >0
```
olleolleolle: Perhaps a space after the operator?

hansy:
This change means `sign_in_count` is a required attribute for the `confirmable` module? Doesn't Devise add errors to the instance if it's already been confirmed? If so, wouldn't it make more sense for the confirmations controller to check for errors on the resource before proceeding to generate headers and/or sign it in?

zachfeldman:
@hansy if this is concerning for you please file a new issue since this was already merged! Thanks

MaicolBen:
@hansy It was fixed in #1064

test/controllers/devise_token_auth/confirmations_controller_test.rb
```diff
@@ -61,6 +61,12 @@ def token_and_client_config_from(body)
         test 'User shoud have the Last checkin filled' do
           assert @resource.last_sign_in_at?
         end
+        
+        test 'user already confirmed' do
+          assert @resource.sign_in_count > 0 do
+            assert expiry == (Time.now + Time.now + 1.second).to_i
```
Evan-M: The `#assert` method from minitest does not yield the block that is given here!
(See [RDoc for `Minitest::Assertions#assert`](http://www.rubydoc.info/gems/minitest/Minitest/Assertions#assert-instance_method))

To further demonstrate this, even with an exception raised from within the block, the test still passes!
~~~ruby
test 'user already confirmed' do
  assert @resource.sign_in_count > 0 do
    assert expiry == (Time.now + Time.now + 1.second).to_i
    raise "Chunky Bacon!"
  end
end
~~~

Evan-M: n/m... I commented here waaay late, and this has been resolved by #1113.

PR #1004, opened 2017-11-03
Allow multiple authentication_keys on login

danielneis:
Hello,

this is second attempt of https://github.com/lynndylanhurley/devise_token_auth/pull/949
I rebased the patch to current master and added back the "binary" statement for mysql


zachfeldman:
Hey @danielneis looks like the build is failing - could you perhaps better explain what your pull request is doing? I read through it but I'm not sure if I understand. Thanks for taking the time!

zachfeldman:
Hey @danielneis looks like the build is failing - could you perhaps better explain what your pull request is doing? I read through it but I'm not sure if I understand. Thanks for taking the time

danielneis:
Hi, @zachfeldman , what I am trying to do is to use all the resource_class.authentication_keys instead of only the first one. The users in my database uses two fields for authentication (account and agency, for example).
With the first patch I was able to get it working on postgresql, I didn't tested this new code, just rewrite it to fit the new structure but it looks like it not worked...

zachfeldman:
Got it @danielneis - well if the checks pass I'd be down to merge, I don't see a huge problem with that.

MaicolBen:
You didn't define `fields` in this PR, maybe it's that. I am not sure about adding this feature, but make sure this doesn't break for the common scenario of one field.

For the next time, please put a PR here when the work is finished.

danielneis:
Now the tests are passing =)

It would be nice to have some tests for those cases with multiple :authentication_keys 
It would have to define it on the model and then try to login with and without those fields, with users that may have same and different values on the other fields.
If you have any ideas or directions on how to do it I can add the tests =)

ArielAleksandrus:
Isn't this branch merged yet? Can you please write a guide on how to use this new feature? Thank you in advance

danielneis:
Hello,

if you have on your devise configuration something like that:

    config.authentication_keys = [:account, :domain]

Then you will have on your User model something like:

    devise  :database_authenticatable, :authentication_keys => [:account, :domain]

This will use the two fields for comparison when logging someone in. The account and domain must match.

With this patch, it will also work with devise_token_auth =)

app/controllers/devise_token_auth/concerns/resource_finder.rb
```diff
@@ -12,14 +12,26 @@ def get_case_insensitive_field_from_resource_params(field)
...
-    if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
-      q = "BINARY " + q
+  def find_resource
+
+    fields = (resource_params.keys.map(&:to_sym) & resource_class.authentication_keys)
+
+    conditions = []
+    values = {}
+    fields.each do |f|
+      q = " #{f.to_s} = :#{f.to_s} "
```
MaicolBen: use more meaningful variables, maybe `query_field` ?

app/controllers/devise_token_auth/concerns/resource_finder.rb
```diff
@@ -12,14 +12,26 @@ def get_case_insensitive_field_from_resource_params(field)
...
+      end
+      conditions.push(q)
+      values[f.to_sym] = get_case_insensitive_field_from_resource_params(f)
     end
 
-    @resource = resource_class.where(q, value).first
+    conditions.push(' provider = :provider')
+    values[:provider] = provider.to_s
+
+    @resource = resource_class.where([conditions.join(" AND "), values]).first
```
MaicolBen: use single quotes

MaicolBen:
Sounds good! thanks, this resolves #1128

app/controllers/devise_token_auth/concerns/resource_finder.rb
danielneis: sure! just did it.

app/controllers/devise_token_auth/concerns/resource_finder.rb
danielneis: sure! just did it. called q = condition and f = field.

danielneis:
I took a look at the current tests and I'm afraid I don't know where to begin.
Maybe it would need a new "user" class with those options?
If you could help with thats would be awesome.

lpsBetty:
@danielneis @MaicolBen will this be resolved anytime soon?

PR #1009, opened 2017-11-07
Update CONTRIBUTING.md

stratigos:
* Updated link to Flavio Castelli blogpost about executing a single test

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -28,4 +28,4 @@ bundle exec guard
 
 Guard will re-run each test suite when changes are made to its corresponding files.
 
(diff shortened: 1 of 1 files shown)


zachfeldman:
Haha I feel like I just merged a PR like this but I'll do it sure.

stratigos:
Yah my guess is that someone noticed the original link expired (author changed the site's TLD), then found a cached google link with the content. The cached google content expired, and I managed to find the author's new site, and subsequently found the original post.

I needed to find it as I havent used the core Ruby test library in years, and am only versed in RSpec (for Ruby tests, that is).

zachfeldman:
For sure!

Issue #887, opened 2017-05-12
API Does Not Use Api_Controller 


iamameme:
Sorry, but I couldn't find a solution to this anywhere. In the documentation, it says that if you are using both devise and devise_token_auth, you need two different controllers.

```
# app/controllers/api_controller.rb
# API routes extend from this controller
class ApiController < ActionController::Base
  include DeviseTokenAuth::Concerns::SetUserByToken
end

# app/controllers/application_controller.rb
# leave this for ActiveAdmin, and any other non-api routes
class ApplicationController < ActionController::Base
end
```

I did this, but the API calls still go to the normal application controller and don't ever reach the API controller. How do you connect them?

zachfeldman:
Hi there @iamameme    ,

In an effort to cleanup this project and prioritize a bit, we're marking issues that haven't had any activity in a while with a "close-in-7-days" label. If we don't hear from you in about a week, we'll be closing this issue. Obviously feel free to re-open it at any time if it's the right time or this was done in error! 

If you are still having the issue (especially if it's a bug report) please refer to our new [Issue Template](https://github.com/lynndylanhurley/devise_token_auth/blob/master/.github/ISSUE_TEMPLATE.md) to provide some more details to help us solve it.

Hope all is well.

RigoTheCEO:
+ 1

RigoTheCEO:
+1

zachfeldman:
@RigoTheCEO would you prefer we keep this issue open?

RigoTheCEO:
Yes. We should be able to decide which application class this gem descends from. This could be helpful in many ways, especially when using the standard Devise gem with this one.

zachfeldman:
Awesome! @RigoTheCEO can you submit a pull request to that effect if you feel this way? I'd be glad to review.

karlingen:
@zachfeldman PR up for grabs!

karlingen:
@iamameme see if the latest master fixes your issue

PR #1016, opened 2017-11-12
Fixes include bug causing sign_in to require auth

### LINKED ISSUE #1014: API is not using defined ApiController but rather ApplicationController
Doing a POST to `/api/v1/auth/sign_in` with email and password causes status 401.
Removing the line `before_action :authenticate_user!` from ApplicationController resolves the issue but breaks rest of the app. 

Using version `v0.1.42` of the gem.

Why is this happening when the auth is not even mounted to ApplicationController?
Also, the `sign_in` method should always ignore `:authenticate_user!` just like Devise. 

Controllers:
```
class Api::ApiController < ActionController::Base
  include DeviseTokenAuth::Concerns::SetUserByToken
end

class ApplicationController < ActionController::Base
  before_action :authenticate_user!
end
```

Routes: 
```
devise_for :users, path: '', path_names: { sign_in: 'login', sign_out: 'logout' }
# API
namespace :api, defaults: { format: 'json' } do
  scope :v1 do
    mount_devise_token_auth_for 'User', at: 'auth', controllers: { sessions: "users/sessions" }
    get :test
  end
end
```

Curl: 
```bash
$ curl -XPOST 'http://localhost:3000/api/v1/auth/sign_in' -d '{ "email":"foo@bar.com", "password":"123456" }' -H "Content-Type: application/json" -H 'Content-Type: application/json'
```

karlingen:
This fixes #1014, #887 and probably some more issues. 

I'm not sure why that line was included in that file in the first place. It is already being loaded inside `DeviseTokenAuth::Concerns::ResourceFinder` which also is included. 

Now I am successfully able to run devise alongside devise_token_auth. 

Leave my ApplicationController untouched:

```ruby
class ApplicationController < ActionController::Base
  protect_from_forgery with: :exception
  before_action :authenticate_user!
end
```

and use another controller for the Api: 

```ruby
class Api::ApiController < ActionController::Base
  include DeviseTokenAuth::Concerns::SetUserByToken
end
```

routes: 
```ruby
namespace :api, defaults: { format: 'json' } do
  scope :v1 do
    mount_devise_token_auth_for 'User', at: 'auth', controllers: { sessions: "users/sessions" }
    resources :test, only: :index
  end
end
```

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -1,7 +1,6 @@
 module DeviseTokenAuth::Concerns::SetUserByToken
   extend ActiveSupport::Concern
   include DeviseTokenAuth::Concerns::ResourceFinder
-  include DeviseTokenAuth::Controllers::Helpers
 
   included do
     before_action :set_request_start


zachfeldman:
All the tests pass so I don't have a huge problem with this. @MaicolBen @lynndylanhurley any issues?

lynndylanhurley:
Thanks @karlingen. I'm not sure why that concern was included in the first place.

naseeratevolverstech:
@karlingen i have tried your solution but it does not work for me. i am using devise for admin and vendor but devise_token_auth for user model making apis for users but when i make auth sign up it shows error invalid authenticity token. and when i wirte protect_from_forgery with: :null_session.in application controller, it signed up successfully and after that when i try to sign in it shows json error user already exits. i am stuck please help me.

Issue #1008, opened 2017-11-07
Reset Password Tokens No Longer Expire


stratigos:
The Devise configuration for: 

```ruby
Devise.setup do |config|
  # ...
  config.reset_password_within = 30.seconds
  #...
end
```
no longer has any effect in DeviseTokenAuth. I believe it has been bypassed by #957. As pointed out in #672, the `DeviseTokenAuth::PasswordsController#edit` was previously not returning an error properly when an expired token was found, which I think is noteworthy given this recently merged PR.

Essentially, the call to `Devise::Models::Recoverable#reset_password_period_valid?` is never made. I will manually add this to an override in my application for now, and will get to work on a PR.

----

Devise handles this in [Devise::PasswordsController#update](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/passwords_controller.rb#L32). The resource is sent [#reset_password_by_token](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L143), which in turn calls [#reset_password_period_valid?](https://github.com/plataformatec/devise/blob/ee01bac8b0b828b3da0d79c46115ba65c433d6c8/lib/devise/models/recoverable.rb#L75). So to me, it looks like the call to `#reset_password_by_token` was removed with #957, thus skipping over some expected functionality, including checking if the token has expired yet or not. 

Im no crypto expert, but I feel that the callback to `resource_class.with_reset_password_token` appears less secure than `resource_class.reset_password_by_token`.

----
**Version**:   `devise_token_auth (0.1.42)`
**Environmental Info**: 
  * **Routes**: scoped to "v1"
  * **Config**: I am testing this by setting the `devise` config for `reset_password_within` to be 30 seconds. Other than that, all that is customized is my mailer account/host info.
  * **Custom Overrides**: nothing in terms of passwords, but have overridden the Sessions controller to use the yield block to accept custom parameters on login.

zachfeldman:
@stratigos good catch if this is really happening. We'd welcome your PR!

stratigos:
Yeah Im going to work on it today/tonight hopefully, if not tomorrow, and am pretty sure it should be easy to restore to the `#update` action.

stratigos:
So, I worked on this for a while, and came to the following: 

[...]

PR #1017, opened 2017-11-13
Restore reset password tokens behavior

dks17:
Please see Reset Password Tokens No Longer Expire #1008


stratigos:
Following the discussion in #1008, this still forces the default use case (i.e., not Mongoid) to make a local override to handle helpful error messages (all fails return a generic 404).

zachfeldman:
These changes make sense to me. Merging

PR #1028, opened 2017-11-23
Honor strip_whitespace_keys from devise config

nerfologist:
Hello,
in our application  some (clumsy?) users are including whitespace before or after their email address when trying to log in, and failing (credentials were being rejected). We are also not stripping the whitespace in the front-end.

Based on our understanding, it seems that `devise_token_auth` is not honoring the `strip_whitespace_keys` as set in `config/initializers/devise.rb`.

This small patch will strip whitespace according to the `strip_whitespace_keys` configuration.

(tested and passing on Ruby 2.4.1)

Hope to help, thanks for this great gem!
Marco

### DIFF
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -9,6 +9,10 @@ module DeviseTokenAuth::Concerns::ResourceFinder
     if resource_class.case_insensitive_keys.include?(field.to_sym)
       q_value.downcase!
     end
+
+    if resource_class.strip_whitespace_keys.include?(field.to_sym)
+      q_value.strip!
+    end
     q_value
   end
 
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -13,6 +13,7 @@ module DeviseTokenAuth::Concerns::ResourceFinder
     if resource_class.strip_whitespace_keys.include?(field.to_sym)
       q_value.strip!
     end
+
     q_value
   end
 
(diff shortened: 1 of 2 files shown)


lynndylanhurley:
This change looks good to me. Any idea why the tests are failing?

I need to figure out how to disable all of those warnings in travis - it's impossible to see what the actual error was through all the noise.

nerfologist:
Hi, I tried to run the test suite on ruby 2.4.2 and all three databases (sqlite, mysql and postgresql, albeit version 10) and it failed the first time, but mysteriously worked on the second and following times. I have no idea why at the moment, so I was looking for a way to restart the build and see if it would fix itself... 😄

lynndylanhurley:
The error may be unrelated to this PR. I'll pull down and test ASAP.

zachfeldman:
closing and reopening to try running the tests again

zachfeldman:
Unfortunately looks like tests are still failing here :(

lynndylanhurley:
Looks like this is the failing test:
```
test_request_should_succeed_if_configured#DeviseTokenAuth::SessionsController::Confirmed user::stripping whitespace
```

@nerfologist are you sure you're setting the class variable correctly?

```
@resource_class.strip_whitespace_keys = [:email]
```

Could it be that you have to do something like this?

```
@resource_class.class_variable_set(:@@strip_whitespace_keys = [:email])
```

nerfologist:
Thanks all for your feedback, I'll look into it ASAP.

nerfologist:
There was a leftover `.upcase` from the nearby `case_insensitive_keys` config option test case (which I copy&pasted). Should be good now.

I can't figure out why it kind of worked for most ruby versions/databases though 🤔 ...

MaicolBen:
It's okay, thanks!

If someone else approves this PR, we'll merge it

PR #719, opened 2016-08-30
Remove params that are not being used in resource params

jahammo2:
#### Why

Just to avoid confusion. When I did some overrides on a project awhile back, it was a little confusing.
#### What

Simply deleted the params not being used in the controller

PR #1029, opened 2017-11-24
Remove params that are not being used in resource params

MaicolBen:
Redone #719, thanks to @jahammo2

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -189,7 +189,7 @@ module DeviseTokenAuth
     private
 
     def resource_params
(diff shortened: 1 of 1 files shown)


zachfeldman:
Makes sense to me. We only need UID of some kind and reset password token to do the work of that endpoint. Merging.

mrkrlli:
If we still allow the user to pass in `redirect_url` as a param for the POST request to `/auth/password`, wouldn't we still need to permit the `redirect_url` parameter? Otherwise an unpermitted parameter error is thrown.

It's a bit confusing, since the docs for the endpoint still say to pass in `redirect_url`, which would be required if the user doesn't set a default in the initializer.

PR #1030, opened 2017-11-24
Fix bypass_sign_in broken tests

MaicolBen:
Fixed tests that #911 broke in master. I should have rebuilt travis for that PR before merge because it had a long time opened, now I learned the lesson


zachfeldman:
Not a huge change and doesn't break anything and from a regular, @MaicolBen . I merged!

Issue #1031, opened 2017-11-26
validate_token route omits specified json format


dks17:
At each request /api/v1/auth/validate_token browser gets response in JSON format, but it is logged like HTML.

rails server log:

> Started GET "/api/v1/auth/validate_token" for 127.0.0.1 at 2017-11
> Processing by DeviseTokenAuth::TokenValidationsController#validate_token **as HTML**

In browser response is returned like JSON

Request headers:
> Accept: application/json, text/plain, */*

Response headers:
> Content-Type: application/json; charset=utf-8

routes.rb
```
scope '/api/v1', defaults: {format: :json} do
  mount_devise_token_auth_for 'User', at: 'auth'
  #
end
```
rake routes
```
         user_confirmation POST     /api/v1/auth/confirmation(.:format)       devise_token_auth/confirmations#create {:format=>:json}
     new_user_confirmation GET      /api/v1/auth/confirmation/new(.:format)   devise_token_auth/confirmations#new {:format=>:json}
                           GET      /api/v1/auth/confirmation(.:format)       devise_token_auth/confirmations#show {:format=>:json}
api_v1_auth_validate_token GET      /api/v1/auth/validate_token(.:format)     devise_token_auth/token_validations#validate_token
                           GET      /api/v1/auth/validate_token(.:format)     devise_token_auth/token_validations#validate_token
                           GET      /api/v1/auth/validate_token(.:format)     devise_token_auth/token_validations#validate_token
                           GET      /api/v1/auth/validate_token(.:format)     devise_token_auth/token_validations#validate_token
       api_v1_auth_failure GET      /api/v1/auth/failure(.:format)            devise_token_auth/omniauth_callbacks#omniauth_failure
                           GET      /api/v1/auth/:provider/callback(.:format) devise_token_auth/omniauth_callbacks#omniauth_success
                           GET|POST /omniauth/:provider/callback(.:format)    devise_token_auth/omniauth_callbacks#redirect_callbacks
          omniauth_failure GET|POST /omniauth/failure(.:format)               devise_token_auth/omniauth_callbacks#omniauth_failure
```

I also noticed some strange path helpers name **api_v1_auth_validate_token** and **api_v1_auth_failure**. It could be unnoticed when routes is written with [namespace](http://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Scoping.html#method-i-namespace).

I don't use omntiauth. May further paths be duplicates for GET method?
```
[...]

PR #1032, opened 2017-11-26
fix default parameters for validate_token

dks17:
Please see #1031. 
It allows to use default parameters for validate_toke path like `{format: :json} `

### DIFF
--- a/lib/devise_token_auth/rails/routes.rb
+++ b/lib/devise_token_auth/rails/routes.rb
@@ -36,6 +36,9 @@ module ActionDispatch::Routing
         # get full url path as if it were namespaced
         full_path = "#{@scope[:path]}/#{opts[:at]}"
 
+        # get default parameters for a route
+        defaults = @scope[:defaults] || {}
+
         # get namespace name
         namespace_name = @scope[:as]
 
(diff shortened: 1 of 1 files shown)


zachfeldman:
PR makes sense to me @dks17 but this breaks the test suite.

dks17:
@zachfeldman reopen please. Maybe Travis is alive again.
Worker information

> hostname: 61dfc363-9fea-4b18-a8a2-851bf23accd8@1.i-08efbd1-production-2-worker-org-ec2.travisci.net
> version: v3.4.0 https://github.com/travis-ci/worker/tree/ce0440bc30c289a49a9b0c21e4e1e6f7d7825101
> instance: f450c2e travisci/ci-garnet:packer-1503972846 (via amqp)
> startup: 975.036831ms
> oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

dks17:
@zachfeldman test passed. I managed with it myself.

zachfeldman:
Nice, I will merge this if one other regular contributor approves.

lynndylanhurley:
@dks17 - just to clarify, is this for defining default options as described [here in the rails docs](http://guides.rubyonrails.org/routing.html#defining-defaults)?

If that's the case, do we need to add this to all of the other routes as well?

lynndylanhurley:
Also we should probably write a test for this. Can you think of a good way to test @dks17 ?

dks17:
@lynndylanhurley 
>  is this for defining default options as described here in the rails docs?

Yes.

> do we need to add this to all of the other routes as well?

As I understand devise routes properly accept defaults. I think we should considering only validate token route and omniauth routes. Also I don't use omniauth workflow and can't answer exactly.

> Can you think of a good way to test ?

For now I will try  to cover only one validate token route.

dks17:
@lynndylanhurley I don't know this is a good way to test or not, but a one of possible ways.
To pass this test #1032 should be merged before.

dks17:
Seems to need a more complex approach for routes mounting with refactoring.

PR #1038, opened 2017-11-29
validate_token route defaults test

dks17:
Test for #1032 
To pass test merge #1032 before.
Also see issue #1031


dks17:
Closed. The tests should be together with code.

PR #1042, opened 2017-12-03
Fix polish translations

sbadura:


zachfeldman:
@sbadura I have no idea how to evaluate this as I don't speak Polish, but can you explain the changes you've made?

sbadura:
@zachfeldman I'm Polish. Current translations are wrong. You can compare them with proper translations from: https://github.com/tigrish/devise-i18n/blob/master/rails/locales/pl.yml

zachfeldman:
Good enough for me!

PR #1044, opened 2017-12-05
Translate into Vietnamese

minhthuan274:
Hi man, I am Thuan. Currently I'm a web developer using Rails and Angular. I usually use your gem to build my projects. It's very useful. But not able to Vietnamese customers. So I translated your gem. So please accept my pull request to make your gem more useful for more people, especially with the Vietnamese. Thank you!
I have passed all unit test.

### DIFF
--- a/config/locales/vi.yml
+++ b/config/locales/vi.yml
@@ -1,0 +1,50 @@
+vi:
+  devise_token_auth:
+    sessions:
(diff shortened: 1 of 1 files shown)


MaicolBen:
Thanks! I checked some of them with Google Translator and they are ok!

Issue #1018, opened 2017-11-13
omniauth-facebook yields ActiveModel::ForbiddenAttributesError in omniauth_callbacks_controller.rb:84:in `assign_provider_attrs'


Blitzkev:
Using latest master commit: 
`gem "devise_token_auth", git: "https://github.com/lynndylanhurley/devise_token_auth", ref: "2a5cedbbd886329a318298d15b38fabbee51e7da"`

After successfully calling /auth/facebook, logging in, and being redirected to /auth/facebook/callback, I receive the following error:
`ActiveModel::ForbiddenAttributesError in DeviseTokenAuth::OmniauthCallbacksController#omniauth_success`

After doing some debugging I was able to trace the origin of the error to this part of the omniauth_controller:
```
From: /usr/local/rvm/gems/ruby-2.4.1/bundler/gems/devise_token_auth-2a5cedbbd886/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb @ line 82 DeviseTokenAuth::OmniauthCallbacksController#assign_provider_attrs:

    81: def assign_provider_attrs(user, auth_hash)
 => 82:   binding.remote_pry
    83:   attrs = auth_hash['info'].slice(*user.attributes.keys)
    84:   user.assign_attributes(attrs)
    85: end

[1] pry(#<DeviseTokenAuth::OmniauthCallbacksController>)> attrs = auth_hash['info'].slice(*user.attributes.keys)
=> {"image"=>
  "http://graph.facebook.com/v2.6/REDACTED/picture",
 "email"=>"blitzkev@gmail.com"}
[2] pry(#<DeviseTokenAuth::OmniauthCallbacksController>)> attrs.permitted?
=> false
```

Do I need to explicitly permit image and email on user somewhere? I feel like I'm missing an important configuration step.

gustavosantiago:
i have the same problem

Rails.application.config.middleware.use OmniAuth::Builder do
  provider :facebook, ENV['FACEBOOK_KEY'], ENV['FACEBOOK_SECRET']
  provider :google_oauth2, ENV['GOOGLE_CLIENT_ID'], ENV['GOOGLE_CLIENT_SECRET']
end`
`

PR #1050, opened 2017-12-23
fixing ActiveModel::ForbiddenAttributesError on oAuth login

ChristianHeimke:
I'm using the gem with rails 5.1.1 in api only mode with https://github.com/neroniaky/angular2-token. It works like a charm - but the oauth made some trouble.

Like in issue #1018 I got the _ActiveModel::ForbiddenAttributesError: ActiveModel::ForbiddenAttributesError_ on every login via oauth (tested with twitter and facebook). 

This PR solved the oauth login for me and I got all things back working.

### DIFF
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -79,8 +79,8 @@ module DeviseTokenAuth
 
     # break out provider attribute assignment for easy method extension
     def assign_provider_attrs(user, auth_hash)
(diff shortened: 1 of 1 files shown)


zachfeldman:
@ChristianHeimke thanks for your contribution! I can't approve this pull request because it modifies the behavior and keeps the method name the same....the method says it will assign attributes which doesn't persist anything to the database while you've updated the method to update attributes instead which will persist.

ChristianHeimke:
@zachfeldman thank you for the feedback, I totally agree with you. I'll close it for now and update it as soon as I got time for it.

zachfeldman:
Happy holidays @ChristianHeimke !

PR #1055, opened 2017-12-30
Separate error messages for missing vs invalid email

pjungwir:
When you submit a missing email, you should only get an error message that it's missing, not that it's missing and that it's an invalid format.


lynndylanhurley:
Thanks @pjungwir!

PR #1061, opened 2018-01-09
Refactor token creation

MaicolBen:
I didn't know if I should include `@resource.save!` in the new method, also method name suggestions are appreciated.

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -4,20 +4,12 @@ module DeviseTokenAuth
       @resource = resource_class.confirm_by_token(params[:confirmation_token])
 
       if @resource && @resource.id
-        # create client id
-        client_id  = SecureRandom.urlsafe_base64(nil, false)
-        token      = SecureRandom.urlsafe_base64(nil, false)
-        token_hash = BCrypt::Password.create(token)
-        expiry     = (Time.now + @resource.token_lifespan).to_i
-
-        if defined? @resource.sign_in_count && @resource.sign_in_count > 0
+        expiry = nil
+        if defined?(@resource.sign_in_count) && @resource.sign_in_count > 0
           expiry = (Time.now + 1.second).to_i
         end
 
-        @resource.tokens[client_id] = {
-          token:  token_hash,
-          expiry: expiry
-        }
+        client_id, token = @resource.create_token expiry: expiry
 
         sign_in(@resource)
         @resource.save!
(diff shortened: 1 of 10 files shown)


KelseyDH:
It might be a good idea to [https://lucaguidi.com/2014/03/27/thread-safety-with-ruby/](avoid the use of `||=`) if we can, [https://stackoverflow.com/questions/9853086/is-in-ruby-thread-safe#9855171](as apparently its not thread safe in MRI Ruby.)

MaicolBen:
Thanks, but that applies to instance variables, you can't share two method variables unless you create a thread inside that method.

zachfeldman:
@MaicolBen sorry I haven't seen this yet! I approve of this line of thought and your changes. Fix your conflicts and we should be good to merge!

Evan-M:
How come the [`#update_auth_header`][`SUBT#update_auth_header`] instance method in the [`SetUserByToken`][`DeviseTokenAuth::Concern::SetUserByToken`] controller concern doesn't utilize this refactoring? 

Instead it uses the [`DeviseTokenAuth::Concerns::User#create_new_auth_token`][`#create_new_auth_token`] instance method, which seems very similar to the [`#create_token`] method added in this PR!

Are these two instance methods in [`DeviseTokenAuth::Concern::User`] ([`#create_new_auth_token`] and [`#create_token`]) duplicate logic? Could they be refactored to DRY up the logic? 

Additionally, I would think that [`#build_auth_header`] and [`#update_auth_header`][`U#update_auth_header`] in the [`DeviseTokenAuth::Concern::User`] model concern, are perhaps not a proper separation of concerns (`MVC`).  Does a `User` model (or other `Devise` resource) need to know about request/response headers? 

As an aside, although functionally there is no difference, per Rails convention, instance methods added to an `ActiveSupport::Concern` should be defined at the module level, rather than inside the `included do` block.  The `included do` block is intended to have associations, validations, serializations, scopes, or other macros set on the instance.

[`DeviseTokenAuth::Concern::SetUserByToken`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb

[`SUBT#update_auth_header`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L144

[`DeviseTokenAuth::Concern::User`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.r

[`#create_token`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L110

[`#create_new_auth_token`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L195

[`#build_auth_header`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L217

[`U#update_auth_header`]: https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L231

PR #1064, opened 2018-01-12
Make confirmable not dependant on trackable

romankovt:
Hello,

At the moment we have confirmable dependant on trackable because it excplicitly calls sign_in_count column while confirming email, but one should have an ability to use it without trackable module

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -10,7 +10,7 @@ module DeviseTokenAuth
         token_hash = BCrypt::Password.create(token)
         expiry     = (Time.now + @resource.token_lifespan).to_i
 
-        if @resource.sign_in_count > 0
+        if defined? @resource.sign_in_count && @resource.sign_in_count > 0
           expiry = (Time.now + 1.second).to_i
         end
 


zachfeldman:
@romankovt I think this change makes sense. If another maintainer approves it, we can merge it.

app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -10,7 +10,7 @@ def show
         token_hash = BCrypt::Password.create(token)
         expiry     = (Time.now + @resource.token_lifespan).to_i
 
-        if @resource.sign_in_count > 0
+        if defined? @resource.sign_in_count && @resource.sign_in_count > 0
```
doubleton: @zachfeldman, @lynndylanhurley Guys, there is a bug:
The parentheses are required here around the first `@resource.sign_in_count`, i.e.:
```
```
Otherwise `defined?` is applied to the whole expression and returns string `"expression"`, which results into assigning expiry to 1 second ahead of now even when `sign_in_count` equals 0.

doubleton: There is a simple snippet to show the difference:
```
irb(main):001:0> defined? some_var
=> nil
irb(main):002:0> defined? some_var && some_var > 0
=> "expression"
irb(main):003:0> defined?(some_var) && some_var > 0
=> nil
```

romankovt: made a PR #1067 fixing this, thank you for report

zachfeldman:
Thanks for your vigilance all!

PR #1067, opened 2018-01-16
Fix checking for presence #sign_in_count on resource in confirmations

romankovt:
There is a bug reported in comments https://github.com/lynndylanhurley/devise_token_auth/pull/1064
Here is a pull request fixing the bug.


doubleton:
Seems it was fixed by #1061

romankovt:
yep, you're right so closing it

zachfeldman:
@romankovt thanks for being proactive!

PR #1071, opened 2018-01-25
Update user template

dkniffin:
I copied the template that [is used by default in devise](https://github.com/plataformatec/devise/blob/master/lib/generators/devise/orm_helpers.rb#L10). The main reason I did this is that it was adding `omniauthable` by default, even if omniauth is not in the Gemfile.


zachfeldman:
@dkniffin I understand and approve of this change, I suppose it's kind of silly to assume everyone wants omniauth. If another contributor approves we'll merge!

dkniffin:
Thanks guys!

PR #1073, opened 2018-01-25
Normalize session errors to align with regristration

stfnndrsn:
There was inconsistency between error messages from session vs. registration validation. 

I've updated the session controller to follow the same pattern as the one used in registration. 

_Invalid password:_
```json
{
    "errors": [
        "Invalid login credentials. Please try again."
    ]
}
```

```json
{
    "errors": {
        "full_messages": [
            "Invalid login credentials. Please try again."
        ]
    },
    "success": false
}
```

(the added `success` key must be a change from the released gem to the master-branch checkout - but it does not match up with registration use of `error`)

**An error in registration:**
```json
{
    "data": {
        "created_at": null,
        "email": "stfnandersen14@gmail.com",
        "id": null,
        "provider": "email",
        "uid": "",
        "updated_at": null
    },
    "errors": {
        "full_messages": [
            "Password can't be blank"
        ],
        "password": [
            "can't be blank"
        ]
    },
    "status": "error"
}

```

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -67,7 +67,9 @@ module DeviseTokenAuth
     def render_error(status, message, data = nil)
       response = {
         success: false,
-        errors: [message]
+        errors: {
+          full_messages: [message]
+          }
       }
       response = response.merge(data) if data
       render json: response, status: status
--- a/test/controllers/devise_token_auth/passwords_controller_test.rb
+++ b/test/controllers/devise_token_auth/passwords_controller_test.rb
@@ -30,7 +30,8 @@ class DeviseTokenAuth::PasswordsControllerTest < ActionController::TestCase
 
         test 'error message should be returned' do
           assert @data['errors']
-          assert_equal @data['errors'],
+          assert @data['errors']['full_messages']
+          assert_equal @data['errors']['full_messages'],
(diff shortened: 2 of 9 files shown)


zachfeldman:
@stfnndrsn thanks for your PR! I agree it's important to have consistency here. However, this is probably a breaking change for anybody updating the gem. Is it possible to have an interim solution that returns the, "normal" messages unless a flag is set, with a deprecation warning, then putting this in in full in the next minor version perhaps?

stfnndrsn:
@zachfeldman Of course; altough I'm not sure how to proceed with this as an configuration option and the framework though

zachfeldman:
`config.check_current_password_before_update` is one example of a config option in the README. I'd do a full text search for that.

Perhaps `config.enable_normalized_session_errors` would work well! Then yeah next time we do a more major update we just enable it by default then sometime in the future, completely remove the mismatched message.

Does that help @stfnndrsn ?

MaicolBen:
I don't agree with this PR, it's weird that you reply with `full_messages` when it isn't an active record's error. In fact, we should change to:
```
{
    "error": "Invalid login credentials. Please try again."
}
```

That's something we can do for the next version or use a config like Zach said.

MaicolBen:
Closing this, we should make a breaking change returning `error: "string message"` instead

PR #1075, opened 2018-01-31
Simplifying ConfirmationsController show behavior

dks17:
Is this snippet is really [needed there](https://github.com/lynndylanhurley/devise_token_auth/blob/952084a9b63dad5f831be4589bc172922f24a4b9/app/controllers/devise_token_auth/confirmations_controller.rb#L7)?

```ruby
expiry = nil
  expiry = (Time.now + 1.second).to_i
end
```
```ruby
module DeviseTokenAuth
  class ConfirmationsController < DeviseTokenAuth::ApplicationController
    def show
      @resource = resource_class.confirm_by_token(params[:confirmation_token])

      if @resource && @resource.id
        expiry = nil
          expiry = (Time.now + 1.second).to_i
        end

        client_id, token = @resource.create_token expiry: expiry

        sign_in(@resource)
        @resource.save!

        redirect_header_options = {account_confirmation_success: true}
        redirect_headers = build_redirect_headers(token,
                                                  client_id,
                                                  redirect_header_options)
        redirect_to(@resource.build_auth_url(params[:redirect_url],
                                             redirect_headers))
      else
        raise ActionController::RoutingError.new('Not Found')
      end
    end
  end
end
```

Should behavior of show action depends on including [Trackable module](https://github.com/plataformatec/devise/blob/bdd6081815ca80705aada4cf77aed3b2a2df71c7/lib/devise/models/trackable.rb#L9) or not?
What if Trackable module is off?
Should that behavior be covered with some tests?


zachfeldman:
Hey @dks17 , perhaps you're right that we shouldn't depend on trackable here. In that case, are you suggesting we set no expiry on the sign in token instead?

dks17:
@zachfeldman I am suggesting do less than more.
Is it necessary to compute conditional block if possible and acceptable doesn't do it?
If expiry is not provided it is [set by default](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/app/models/devise_token_auth/concerns/user.rb#L113)
```ruby
expiry ||= (Time.now + token_lifespan).to_i
```
I just slimplified it much as were possible. Of course we could leave only
```ruby
expiry = (Time.now + 1.second).to_i
client_id, token = @resource.create_token expiry: expiry
```
I think conditional block should be cleaned up anyway.

zachfeldman:
Ok, I think this change makes sense then. If another maintainer approves this pull request, I will merge it.

dks17:
I would like implement TokenFactory to generate token in one place with opportunity of tuning token and I have almost done it.
I would like that this pr will be considered soon. If this pr will be rejected I will add to my next pr a code to which we will come. Or I will just close this pr and forget about it.
I think this code should be refactored and simplified:
```ruby
expiry = nil
 expiry = (Time.now + 1.second).to_i
 end

client_id, token = @resource.create_token expiry: expiry
```
@lynndylanhurley, @MaicolBen 
what do you think about changes?

MaicolBen:
I don't quite understand why you're removing the code, it was added by #519 (I remade the PR in #1001). 
It makes sense that the tests are still working without the expiration change, but it doesn't make sense your argument "Should behavior of show action depends on including Trackable module or not?". We try to support all the device modules, and if there are a specific module code, it should ask for the module presence (in this case asks for the presence of `sign_in_count` definition, so it's the same).

MaicolBen:
The idea of #519 is to allow only one confirmation, it expires the token with `expiry = (Time.now + 1.second).to_i`

dks17:
Where is place I could turn trackable off in the gem? Is it [there](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/app/models/devise_token_auth/concerns/user.rb#L21)?
I would like check this statement:

> [You may not want all of these features enabled in your app. That's OK! You can mix and match to suit your own unique style.](https://github.com/lynndylanhurley/devise_token_auth#excluding-modules)

Devise implementation [doesn't use trackable sign_in_count.](https://github.com/plataformatec/devise/blob/bdd6081815ca80705aada4cf77aed3b2a2df71c7/app/controllers/devise/confirmations_controller.rb#L22)

Of course I can be mistaken.

dks17:
I have removed :trackable [from here](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/app/models/devise_token_auth/concerns/user.rb#L21)

```ruby
included do
    # Hack to check if devise is already enabled
    unless self.method_defined?(:devise_modules)
      devise :database_authenticatable, :registerable,
          :recoverable, :validatable, :confirmable
else
```

and got bunch of 9 failures:
```console
FAIL["test_User_shoud_have_the_Last_checkin_filled", Minitest::Result, 16.495720947001246]
 test_User_shoud_have_the_Last_checkin_filled#Minitest::Result (16.50s)
        Expected false to be truthy.

 FAIL["test_User_shoud_have_the_signed_in_info_filled", Minitest::Result, 16.618372673998238]
 test_User_shoud_have_the_signed_in_info_filled#Minitest::Result (16.62s)
        Expected false to be truthy.

 FAIL["test_user_already_confirmed", Minitest::Result, 16.850464181996358]
 test_user_already_confirmed#Minitest::Result (16.85s)
        Expected false to be truthy.

 FAIL["test_the_sign_in_count_should_be_1", Minitest::Result, 17.10246518799977]
 test_the_sign_in_count_should_be_1#Minitest::Result (17.10s)
        Expected false to be truthy.

 FAIL["test_sign_in_count_incrementns", Minitest::Result, 29.07351241700235]
 test_sign_in_count_incrementns#Minitest::Result (29.07s)
        Expected: 1
          Actual: 0

 FAIL["test_current_sign_in_at_is_updated", Minitest::Result, 29.27919944599853]
 test_current_sign_in_at_is_updated#Minitest::Result (29.28s)
        Expected nil to be truthy.

 FAIL["test_last_sign_in_ip_is_updated", Minitest::Result, 29.481255300001067]
 test_last_sign_in_ip_is_updated#Minitest::Result (29.48s)
        Expected: "0.0.0.0"
          Actual: nil

 FAIL["test_sign_in_ip_is_updated", Minitest::Result, 29.64863184800197]
 test_sign_in_ip_is_updated#Minitest::Result (29.65s)
        Expected: "0.0.0.0"
          Actual: nil

 FAIL["test_last_sign_in_at_is_updated", Minitest::Result, 29.857282397002564]
 test_last_sign_in_at_is_updated#Minitest::Result (29.86s)
        Expected nil to be truthy.

409 tests, 562 assertions, 9 failures, 0 errors, 0 skips
```

MaicolBen:
> Devise implementation doesn't use trackable sign_in_count.

I don't know why you pointed confirmations controller, of course, devise doesn't use it there, but we use it here to avoid double confirmation. So we can't turn off the module, because we use it.

You have to turn it off in your code, but seems you have `sign_in_count` defined, so you should remove that field if you aren't interested in tracking the sign in.

dks17:
DeviseTokenAuth includes modules if devise macro(?) with modules is not defined. 
As I understand you can't use a module by default for DeviseTokenAuth if user doesn't include specific module into user model.
You can reproduce the same failures another way. Put it in [https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/dummy/app/models/user.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/dummy/app/models/user.rb)
```ruby
  devise :database_authenticatable, :registerable,
          :recoverable, :validatable, :confirmable
```
This is the same I have done above. I think we should pay attention to the failures.

I think Trackable module is should be optional, for statistics gathering only. And implementation should be independent of the module.

What will happen if Trackable was been off? Should 'If' statement computing thousand/million times even if user didn't include Trackable module?
I think implementation should be the same in the both cases (Trackable off/on).

MaicolBen:
> I think Trackable module is should be optional, for statistics gathering only. And implementation should be independent of the module.
Sorry if I wasn't clear but in this case `defined?(@resource.sign_in_count)` is the same as `resource_class.devise_modules.include?(:trackable)`. So if you don't have enabled trackable in your project, it won't enter to that condition.  

> What will happen if Trackable was been off? Should 'If' statement computing thousand/million times even if user didn't include Trackable module?
I think `defined?(@resource.sign_in_count)` is way faster than `resource_class.devise_modules.include?(:trackable)`. In case you don't want that code at all in your project because it's slow (I doubt it), you can override the show action in your project.

> I think implementation should be the same in the both cases (Trackable off/on).
In this part I can agree, how would you achieve that?

dks17:
> In this part I can agree, how would you achieve that?

Devise [confirmed?](https://github.com/plataformatec/devise/blob/bdd6081815ca80705aada4cf77aed3b2a2df71c7/lib/devise/models/confirmable.rb#L42) method?

```ruby
# expiry = nil
# if defined?(@resource.sign_in_count) && @resource.sign_in_count > 0
#   expiry = (Time.now + 1.second).to_i
# end
expiry = @resource.confirmed? ? (Time.now + 1.second).to_i : nil

client_id, token = @resource.create_token expiry: expiry
```

MaicolBen:
@dks17 Sounds good! Can you change that in the PR? It shouldn't break any test

dks17:
I have caught:

```console
ERROR["test_yield_resource_to_block_on_show_success", Minitest::Result, 26.821582222997677]
 test_yield_resource_to_block_on_show_success#Minitest::Result (26.82s)
NoMethodError:         NoMethodError: undefined method `[]' for nil:NilClass
            app/controllers/custom/confirmations_controller.rb:4:in `show'

ERROR["test_should_redirect_to_success_url", Minitest::Result, 34.8219258479985]
 test_should_redirect_to_success_url#Minitest::Result (34.82s)
NoMethodError:         NoMethodError: undefined method `[]' for nil:NilClass
            

ERROR["test_user_should_now_be_confirmed", Minitest::Result, 38.842425470997114]
 test_user_should_now_be_confirmed#Minitest::Result (38.84s)
NoMethodError:         NoMethodError: undefined method `[]' for nil:NilClass
```
Tests is running randomly. You can catch more errors.
I think new pr is needed for this changes (controller + tests). I would not like including this in TokenFactory pr.

Evan-M:
@dks17: I've run into this same issue while working on a PR for #1123.  I have rewritten `/test/controllers/devise_token_auth/confirmations_controller_test.rb` nearly entirely, but I'm still "randomly" getting the `NoMethodError: undefined method `[]' for nil:NilClass` errors that you mentioned above.

These errors appear to be the result of a race-condition! 

Using the `Timecop` gem, I can manually replicate the two branches of the race condition if I freeze `Time.zone.now` in the tests, and manually advance `Time.zone.now` to either the same second as the auth_token expiry or the second after the auth_token expiry.  To see the race condition, the manual advance in time must happen _after_ generating a new auth_token, but _before_ the resource is saved (with that auth_token).

Essentially, since the expiry is hard-coded to `1.second.from_now`, if either the subsequent calls to `sign_in(@resource)` or `@resource.save!` happen _after_ the expiry, then the newly generated token is deleted by the `DeviseTokenAuth::Concerns::User#destroy_expired_tokens` `before_save` callback.

I've reviewed the changes added in #519 (and #1001), and noticed two things:
1. the call to `sign_in(@resource)` seems to exist to increment the `sign_in_count` (which shouldn't really be used in determining confirmations anyway). [Commit `#15ee641`](MaicolBen/devise_token_auth/commit/15ee641bd7ab479173b0de8bae802bf26441c4a0) seems to imply this as well, but notes that successful confirmation will automatically sign in the user (so the columns _should_ be updated to reflect that)
2. since the `sign_in(@resource)` call updates the `:trackable` columns on the `@resource`, it saves the resource which in turn triggers the `before_save :destroy_expired_tokens` callback. Calling `@resource.save!` will **also** trigger the `before_save :destroy_expired_tokens` callback.  This means there are _two_ opportunities for enough time to pass that a newly created auth_token will have expired (and get destroyed) prior to the success redirect happening.

@maicolben: What is not clear to me is: if the short expiration is conditionally set when a `@resource` has already been confirmed, what is the purpose of setting the expiration to `Time.zone.now + 1.second`, as opposed to `Time.zone.now`? I'm guessing that was an attempt to prevent the `before_save :destroy_expired_tokens` callback from removing the new auth_token prior to redirect?

As I've mentioned in #1123, the confirmations controller _should_ check for the `confirmed_at` attribute to determine if the resource has already been confirmed (instead of `sign_in_count`). However, even an explicit check of `confirmed_at` (or its predicate wrapper-method`confirmed?`) is unnecessary;  `Devise::Models::Confirmable::ClassMethods#confirm_by_token` checks the `confirmed_at` attribute internally, and sets errors on the resource if either not found or already confirmed.

MaicolBen:
@Evan-M I thought we agreed on #1123 to drop `sign_in_count` check, can you make a PR? or do you want me to make one?

Evan-M:
@MaicolBen: I'm already on it. 👍 

However, I've run into a few side effects of that change, and I wanted to ascertain the specifics of the intended behavior when a user clicks on an email confirmation link for a user with an already confirmed email.

jacferreira:
Just wanted to add one more thing regarding `show#confirmations`:

```
  expiry = (Time.zone.now + 1.second).to_i
end
```

In my case I have the default token lifespan (2 weeks) and max number of devices (10). If a user reaches the max number of devices in less than 2 weeks (not at all impossible), the token that is generated on `show#confirmations` will be removed by `clean_old_tokens` since it will be the oldest among the existing tokens. This makes it impossible to be successfully redirected to an authenticated route.

dks17:
Updated. Please see top message.

Issue #1080, opened 2018-02-03
codeclimate-test-reporter soon be deprecated


dks17:
Just for information. [codeclimate-test-reporter - [SOON TO BE DEPRECATED]](https://github.com/codeclimate/ruby-test-reporter)
They suggest using [new test reporter](https://docs.codeclimate.com/v1.0/docs/configuring-test-coverage).

I am not sure but Gemfile should be corrected from:
```ruby
# code coverage, metrics
group :test do
  gem 'codeclimate-test-reporter', require: nil
  gem 'rails-controller-testing'
end
```
to
```ruby
# code coverage, metrics
group :test do
  gem 'simplecov', require: false
  gem 'rails-controller-testing'
end
```

zachfeldman:
@dks17 makes sense. Want to make a PR for this?

dks17:
@zachfeldman Ok. I'll try.

dks17:
@zachfeldman, @MaicolBen 
I hope example 2 will be helpfully.
https://github.com/codeclimate/test-reporter/blob/master/examples/ruby_examples.md

PR #1081, opened 2018-02-03
replace codeclimate-test-reporter gem by simplecov gem

dks17:
Please see #1080


zachfeldman:
@MaicolBen are you ok with this change?

dks17:
@MaicolBen nice catch. @zachfeldman this pr should be closed.

Issue #656, opened 2016-06-04
[Suggestion] Performance issue


yury-dymov:
Hello! First of all thank you for the great library!

I faced a performance issue recently and after profiling figured out that it mostly related to BCrypt.

Please refer https://github.com/codahale/bcrypt-ruby 'Cost Factors' section. Tuning default cost value from 10 to 6 improved performance for creating token_hash from 0.08s to 0.04s on my environment for each request. I went further and set it 1 and now it is better than 0.01s but I am not sure that security-wise it will work good for everybody but anyway default value for token authorization is huge overhead.

Kind regards,
Yuri.

dks17:
Good suggestion.
I think encryption level should be the same for both passwords and tokens because of they both are stored in one database, closely to each other. They should have the same cryptographic strength.

But @yury-dymov  you are right. Hight cost value reduces performance and it is redundantly.
I think the best way it is the cost value should depends on token life span. The shorter life span the lower cost value. For example password is stored long time (years), but token expires in 2 weeks after they are issued.

I noticed that [devise raised up](https://github.com/plataformatec/devise/blob/451ba5362b84ff18a31ea1b3251137128a6b1beb/lib/generators/templates/devise.rb#L100) to 11 default cost value.

I would like that [bcrypt](https://github.com/codahale/bcrypt-ruby/blob/master/lib/bcrypt/engine.rb) token generator cost value depends on token_lifespan period (figures are approximate):
```
token_lifespan          cost
< 2weeks                   4 
> 2weeks and < 1month      8 
1month <                  11
```
This would increase performance in production without serious security consequences.
Or should allow to users set configuration parameter like token_cost independently?

lynndylanhurley:
@dks17 - just to clarify, by changing [the initializer setting](https://github.com/plataformatec/devise/blob/451ba5362b84ff18a31ea1b3251137128a6b1beb/lib/generators/templates/devise.rb#L100) that you linked to, does that update the value used for this gem? Or is this something we would need to build support for?

dks17:
@lynndylanhurley - it uses like default devise config, and I think that it should be overridden. But this relates only user password. I will create pr for this.

PR #1085, opened 2018-02-08
new TokenFactory module

dks17:
Hello!

Now you can create new tokens easily and do it in any place of your application; code for generating tokens is in one place. Also `TokenFactory` provides few auxiliary methods. Other details are discovered below.

**Example:**
```ruby
token = DeviseTokenAuth::TokenFactory.create
#<struct DeviseTokenAuth::TokenFactory::Token client_id="G8IuSUUPnQZdupRwBhFU2w", token="vOHeUtLKj3Y5rQS6NHhoQQ", token_hash="$2a$10$RwsqMnz7twQlMEACI8vYPeuuJYqByt6yZb82BQfzpCn/cEcB6NHz.", expiry=1519306978>

DeviseTokenAuth::TokenFactory.valid_token_hash?(token.token_hash) # =>  true
DeviseTokenAuth::TokenFactory.token_hash_is_token?(token.token_hash, token.token) # => true
```

**1. TokenFactory**

```ruby
require 'bcrypt'

module DeviseTokenAuth
  # A token management factory which allow generate token objects and check them.
  module TokenFactory
    # For BCrypt::Password class see:
    # https://github.com/codahale/bcrypt-ruby/blob/master/lib/bcrypt/password.rb

    # Creates a token instance. Takes an optional lifespan and cost options.
    # Example:
    #   DeviseTokenAuth::TokenFactory.create
    #   => #<struct DeviseTokenAuth::TokenFactory::Token client_id="tElcgkdZ7f9XEa0unZhrYQ", token="rAMcWOs0-mGHFMnIgJD2cA", token_hash="$2a$10$wrsdlHVRGlYW11wfImxU..jr0Ux3bHo/qbXcSfgp8zmvVUNHosita", expiry=1518982690>
    #
    #   DeviseTokenAuth::TokenFactory.create(lifespan: 10, cost: 4)
    #   => #<struct DeviseTokenAuth::TokenFactory::Token client_id="5qleT7_t9JPVcX9xmxkVYA", token="RBXX43u4xXNSO-fr2N_4pA", token_hash="$2a$04$9gpCaoFbu2dUKxU3qiTgluHX7jj9UzS.jq1QW0EkQmoaxARo1WxTy", expiry=1517773268>
    def self.create lifespan: nil, cost: nil
      obj_client_id  = client_id
      obj_token      = token
      obj_token_hash = token_hash(obj_token, cost)
      obj_expiry     = expiry(lifespan)

      Token.new(obj_client_id, obj_token, obj_token_hash, obj_expiry)
    end

    # Generates a random URL-safe client_id.
    # Example:
    #   DeviseTokenAuth::TokenFactory.client_id
    #   => "zNf0pNP5iGfuBItZJGCseQ"
    def self.client_id
      secure_string
    end

    # Generates a random URL-safe token.
    # Example:
    #   DeviseTokenAuth::TokenFactory.token
    #   => "6Bqs4K9x8ChLmZogvruF3A"
    def self.token
      secure_string
    end

    # Returns token hash for a token with given cost. If no cost value is specified,
    # the default value is used. The possible cost value is within range from 4 to 31.
    # It is recommended to not use a value more than 10.
    # Example:
    #   DeviseTokenAuth::TokenFactory.token_hash("_qxAxmc-biQLiYRHsmwd5Q")
    #   => "$2a$10$6/cTAtQ3CBLfpkeHW7dlt.PD2aVCbFRN5vDDJUUhGsZ6pzYFlh4Me"
    #   
    #   DeviseTokenAuth::TokenFactory.token_hash("_qxAxmc-biQLiYRHsmwd5Q", 4)
    #   => "$2a$04$RkIrosbdRtuet2eUk3si8eS4ufeNpiPc/rSSsfpniRK8ogM5YFOWS"
    def self.token_hash token, cost=nil
      cost ||= DeviseTokenAuth.token_cost
      BCrypt::Password.create(token, cost: cost)
    end

    # Returns the value of time as an integer number of seconds. Takes one argument.
    # Example:
    #   DeviseTokenAuth::TokenFactory.expiry
    #   => 1518983359
    #   DeviseTokenAuth::TokenFactory.expiry(10)
    #   => 1517773781
    def self.expiry lifespan=nil
      lifespan ||= DeviseTokenAuth.token_lifespan
      (Time.now + lifespan).to_i
    end

    # Generates a random URL-safe string.
    # Example:
    #   DeviseTokenAuth::TokenFactory.secure_string
    #   => "ADBoIaqXsEDnxIpOuumrTA"
    def self.secure_string
      # https://ruby-doc.org/stdlib-2.5.0/libdoc/securerandom/rdoc/Random/Formatter.html#method-i-urlsafe_base64
      SecureRandom.urlsafe_base64
    end

    # Returns true if token hash is a valid token hash.
    # Example:
    #   token_hash = "$2a$10$ArjX0tskRIa5Z/Tmapy59OCiAXLStfhrCiaDz.8fCb6hnX1gJ0p/2"
    #   DeviseTokenAuth::TokenFactory.valid_token_hash?(token_hash)
    #   => true
    def self.valid_token_hash? token_hash
      !!BCrypt::Password.valid_hash?(token_hash)
    end

    # Compares a potential token against the token hash. Returns true if the token is the original token, false otherwise.
    # Example:
    #   token = "4wZ9gcc900rMQD1McpcSNA"
    #   token_hash = "$2a$10$ArjX0tskRIa5Z/Tmapy59OCiAXLStfhrCiaDz.8fCb6hnX1gJ0p/2"
    #   DeviseTokenAuth::TokenFactory.token_hash_is_token?(token_hash, token)
    #   => true
    def self.token_hash_is_token? token_hash, token
      BCrypt::Password.new(token_hash).is_password?(token)
      rescue
        false
    end

    # Creates a token instance with instance variables equal nil.
    # Example:
    #   DeviseTokenAuth::TokenFactory.new
    #   => #<struct DeviseTokenAuth::TokenFactory::Token client_id=nil, token=nil, token_hash=nil, expiry=nil>
    def self.new
      Token.new
    end

    Token = Struct.new(:client_id, :token, :token_hash, :expiry) do
      # Sets all instance variables of the token to nil. It is faster than creating new empty token.
      # Example:
      #   token.clear!
      #   => true
      #   token 
      #   => #<struct DeviseTokenAuth::TokenFactory::Token client_id=nil, token=nil, token_hash=nil, expiry=nil>
      def clear!
        self.size.times { |i| self[i] = nil }
        true
      end
    end
  end
end
```
As you can see the code of `TokenFactory` is provided with examples. Let's take a look another changes.

**2. Order of Modules**

Order of modules should be:
1. Third-party dependencies.
2. The main module.
2. Own modules.

```ruby
require "devise"

module DeviseTokenAuth
end

require "devise_token_auth/engine"
require "devise_token_auth/controllers/helpers"
require "devise_token_auth/controllers/url_helpers"
require "devise_token_auth/url"
require "devise_token_auth/token_factory"
```

**3. Token Cost**
As you notice now the value of token cost is changable. If cost is not provided `BCrypt` uses [default cost that is 10](https://github.com/codahale/bcrypt-ruby/blob/f2db689ba567d1a018f4dacb1ccf909adc93e887/lib/bcrypt/engine.rb#L5). `DeviseTokenAuth` used this value. In this way this value is left as default value for `token_cost` parameter if no another specified. Behavior of `DeviseTokenAuth` and our expectations have not changed.

`devise_token_auth/lib/devise_token_auth/engine.rb`
```ruby
  self.token_cost = 10
```

For testing default value of token cost is set to 4 which is minimum value of `token_cost`. For current projects `DeviseTokenAuth` initialization file should be updated. The next code should be added:
```ruby
config.token_cost = Rails.env.test? ? 4 : 10
```
On a PC tests were passed in `00:01:18` before `TokenFactory` implementation were made, but after tests are passed in `00:00:15`. Wow!

Thanks @yury-dymov for his [suggestion](https://github.com/lynndylanhurley/devise_token_auth/issues/656).

**4. Need for Speed**
To understand how server load grows in dependence of cost and compare time for each token cost look at further benchmark. Attention! This benchmark will be passed in about 1 hour.
```console
require 'benchmark'

n = 10_000
range = 4..11
token = DeviseTokenAuth::TokenFactory.token

Benchmark.bm do |x|
  range.each do |cost|
    x.report("cost: #{cost}") do
      n.times { DeviseTokenAuth::TokenFactory.token_hash(token, cost) }
    end
  end
end
                user     system      total        real
cost: 4    12.090000   0.000000  12.090000 ( 12.094996)
cost: 5    32.660000   0.020000  32.680000 ( 32.696124)
cost: 6    64.500000   0.000000  64.500000 ( 64.525920)
cost: 7   126.130000   0.020000 126.150000 (126.199861)
cost: 8   250.230000   0.020000 250.250000 (250.334909)
cost: 9   496.750000   0.020000 496.770000 (496.930617)
cost: 10  991.610000   0.070000 991.680000 (991.979662)
cost: 11 1981.270000   0.130000 1981.400000 (1981.973634)
```

For generating a random URL-safe string TokenFacotry uses `SecureRandom.urlsafe_base64`. It uses [64 different symbols](https://ruby-doc.org/stdlib-2.5.0/libdoc/securerandom/rdoc/Random/Formatter.html#method-i-urlsafe_base64) (A-Z, a-z, 0-9, "-"" and "_") to generate a string which length is 22 symbols by default.

Look at Table 1 on page 10 of [A Future-Adaptable Password Scheme](https://www.usenix.org/legacy/events/usenix99/provos/provos.pdf). There is provided some data about a password length, a character set, and password searching time. The most accectable case is in column '62 mixed case alphanumeric' of Table 1. To search a password with length of 9 and compiled from 62 character set is needed about 1786 years.

Some calculations(password length ** number of character set):
```
 9**62          =                           145557834293068928043467566190278008218249525830565939618481
22**64          = 82234029734625049113929522866489193994903509482752292106926951132169536346070980755456
22**62          =   169905020112861671722994881955556185939883284055273330799435849446631273442295414784
22**64 - 22**62 = 82064124714512187442206527984533637808963626198697018776127515282722905072628685340672
```
Notice, that 2 symbols got many additional combinations. Keep in mind that token is generating random strings which are not such a predictable as user passwords.

Let's gather all information in one place:

| Characteristic | Token | Password* |
|---|---|---|
| length | 22 | 9 |
| combinations| 22*64 | 9**62|
| expire | 2 weeks | never |
| hardware | fast | slow |
| searching | ? | 1786 years |

*data from Table 1 of [the scheme](https://www.usenix.org/legacy/events/usenix99/provos/provos.pdf)

Do you still think that a 22 length token that is expired in 2 weeks must be encrypted as a weak password that is stored eternity?

Reduce token `cost` when create a token with short `lifespan`.

**5. new vs clear!**
If you need to get a `Token` object to be empty you should choice `clear!` instead of creating new object. Carry on using the same `Token` object.

```ruby
token = DeviseTokenAuth::TokenFactory.create
token = DeviseTokenAuth::TokenFactory.new

# or

token = DeviseTokenAuth::TokenFactory.create
token.clear!
```

```console
require 'benchmark'

n = 10_000_000
tokens = []
token = DeviseTokenAuth::TokenFactory.create(cost: 4)
n.times { tokens << token.dup;nil }

Benchmark.bm do |x|

  tokens1 = tokens.dup
  x.report("new") do 
    tokens1.each do |token|
      token = DeviseTokenAuth::TokenFactory.new
    end
  end

  tokens2 = tokens.dup
  x.report("clear!") do
    tokens2.each do |token|
      token.clear!
    end
  end
end; nil

            user     system      total        real
new     7.200000   0.000000   7.200000 (  7.201486)
clear!  4.680000   0.000000   4.680000 (  4.686113)
```

**6. Deprecations**
Tests and other code are not touched to show that with `TokenFactory` all test are passed.

There is only one. The method `token_lifespan` and the tests that covers it. `TokenFactory` methods accept `lifespan` parameter directly.
[app/models/devise_token_auth/concerns/user.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/app/models/devise_token_auth/concerns/user.rb#L270)
```ruby
def token_lifespan
  DeviseTokenAuth.token_lifespan
end
```

**7. Override it totally!**
You can override any method of `TokenFactory`. You can even implement any crypto algorithm you like to generate token hash or just generate a token the way you like.

```ruby
module DeviseTokenAuth
  module TokenFactory

    def self.client_id
      # https://ruby-doc.org/stdlib-2.5.0/libdoc/securerandom/rdoc/Random/Formatter.html#method-i-uuid
      SecureRandom.uuid
    end

  end
end

token = DeviseTokenAuth::TokenFactory.create
token.client_id # => "9c6fa3ae-e54d-4c74-b5f7-bf1345880190"
```

**8. Summary**
Now it is only `TokenFactory` can generate tokens and provides couple methods to validate some token's values. And only `TokenFacotry` knows how `Token` works. It returns `Struct` object and if necessary can be replaced by a class. Also some new methdods may be added and current methods could be overriden. `TokenFactory` module should considering as a some base for further improvements of `DeviseTokenAuth`.

Thanks for reading!

**UPDATE:**
- added `Token` instance object method `#present?`
```ruby
      # Checks token attribute presence
      # Example:
      #   token.present?
      #   => true
      def present?
        token.present?
      end
```
- added `bcrypt` gem as the gem dependency
- raised up `mongoid` versions
- generated default initialization `devise.rb` file for `dummy` test application


app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -68,17 +68,17 @@ def set_user_by_token(mapping=nil)
...
     # ensure we clear the client_id
-    if !@token
-      @client_id = nil
+    if !@token.token
+      @token.client_id = nil
       return
     end
 
-    return false unless @token
+    return false unless @token.token
```
Evan-M: Will this line ever get evaluated? isn't `if !@token.token` the same as `unless @token.token`?

Unless I'm totally missing something, I think this line can be refactored into the conditional check above.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -68,17 +68,17 @@ def set_user_by_token(mapping=nil)
     return @resource if @resource && @resource.is_a?(rc)
 
     # ensure we clear the client_id
-    if !@token
-      @client_id = nil
+    if !@token.token
```
Evan-M: Calling `@token.token` seems like a 'smell'... wouldn't it be more idiomatic ruby to say something like `@token` or `@token.present?`?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
dks17: You are on a right way.
I just implemented TokenFactory. You ask me about legacy code. I was trying not touch code to show that all works properly. I think that code refactoring is the next steps, but is not approached for current pr.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
dks17: You are right.
`@token.present?` looks good.
But what should be checked of token object? client_id, token_hash, token, or combinations of them, or all together?
Could it wait for further pull requests? With `TokenFactory` implementation it will be easy (I hope).

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
Evan-M: Yeah, I hear you on holding off refactoring. I'm having similar challenges with the multiple auth provider feature.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: I think you should define the `present?` method with only the token so we don't change the behavior.

MaicolBen:
@dks17 It's a pity losing your work here, do you plan to rebase it or should we close it?

dks17:
@MaicolBen Ok. I will rebase it.

MaicolBen:
@dks17 I guess you want mongo merged before this one, 1.0.0 is out so I can merge these things

dks17:
@MaicolBen  Ok. I will rebased it after the mongoid support #1209 is merged.

dks17:
@MaicolBen It's rebased.
It has been also updated. Please see **UPDATE** section at the top message.

Gemfile.lock
```diff
@@ -1,354 +0,0 @@
-GIT
```
MaicolBen: I thought this already disappeared in the mongo PR but I guess your forgot doing it there

app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -6,21 +6,23 @@ def show
       @resource = resource_class.confirm_by_token(params[:confirmation_token])
 
       if @resource.persisted?
-        expiry = nil
+        lifespan = nil
```
MaicolBen: don't need to init these variables like js, ruby will take it from the scope

app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -6,21 +6,23 @@ def show
...
+        cost     = nil
         if defined?(@resource.sign_in_count) && @resource.sign_in_count > 0
-          expiry = (Time.zone.now + 1.second).to_i
+          lifespan = 1 # 1 second
+          cost     = 4 # 4 is the minimum cost
```
MaicolBen: this part might be better (with the token creation) in another method because this action is getting big

app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
```diff
@@ -156,19 +156,19 @@ def set_random_password
 
     def create_auth_params
       @auth_params = {
-        auth_token:     @token,
-        client_id: @client_id,
-        uid:       @resource.uid,
-        expiry:    @expiry,
-        config:    @config
+        auth_token: @token.token,
```
MaicolBen: I'm thinking that you can create a method in the token factory `auth_params`, but I don't if would be so useful

test/dummy/config/initializers/devise.rb
```diff
@@ -1,17 +1,290 @@
 # frozen_string_literal: true
 
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
-  config.authentication_keys = [:email, :nickname]
+  # The secret key used by Devise. Devise uses this key to generate
```
MaicolBen: do we ned to change this file? I don't know if we need to add this documentation (I know it's in the default devise initializer )

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -1,15 +1,13 @@
 # frozen_string_literal: true
 
-require 'bcrypt'
```
MaicolBen: user shouldn't care about encryption, so 💪

Gemfile.lock
dks17: You are right. I had recovered `Gemfile.lock` before merged it. By bad.

test/dummy/config/initializers/devise.rb
dks17: I think we need. It is only for test application (play with settings easily while testing).

app/controllers/devise_token_auth/confirmations_controller.rb
dks17: Good remark.

app/controllers/devise_token_auth/confirmations_controller.rb
dks17: I think we should hold off refactoring in favour of #1075 or something better.

app/models/devise_token_auth/concerns/user.rb
dks17: What do you mean?

app/controllers/devise_token_auth/confirmations_controller.rb
MaicolBen: Sorry, I thought that PR was dead, ignore this comment

app/models/devise_token_auth/concerns/user.rb
MaicolBen: this is a positive comment, don't need to anything about. I meant that you removed the encryption/tokenization from the user model which follows SRP

dks17:
@MaicolBen Fixed and rebased. Looks ready to merge.

dks17:
@MaicolBen Done. Feel the performance. At least in the tests.

Issue #1084, opened 2018-02-06
ConfirmationsController redirection error


dks17:
If redirect_url parameter is not passed the error is raised by [redirect_to](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/app/controllers/devise_token_auth/confirmations_controller.rb#L23).

example link in email:
`http://localhost:3000/api/v1/auth/confirmation?config=default&confirmation_token=sVvj7xK77XXvdLpvcEbo&redirect_url=`

[DeviseTokenAuth::Url.generate](https://github.com/lynndylanhurley/devise_token_auth/blob/cf55a71398dbc85ef86656abb70c9bcc8ae40ff4/lib/devise_token_auth/url.rb#L13) returns invalid url.

`"://?access-token=m9Pcu5KLiCHQqS9caSR4MQ&account_confirmation_success=true&client=qe7UoBipfgWbKv0B3o8o5g&client_id=qe7UoBipfgWbKv0B3o8o5g&config=default&expiry=&token=m9Pcu5KLiCHQqS9caSR4MQ&uid=koskoskos%40kos.kos"`

```console
ERROR URI::InvalidURIError: bad URI(is not URI?): http://localhost:3000://''?access-token=RiVKf9BooqHGfZ-xxDnLRQ&account_confirmation_success=true&client=pRDv4rkj_S1ooaClrIAe_A&client_id=pRDv4rkj_S1ooaClrIAe_A&config=default&expiry=&token=RiVKf9BooqHGfZ-xxDnLRQ&uid=koskoskos%40kos.kos
	/home/kostya/.rbenv/versions/2.3.3/lib/ruby/2.3.0/uri/generic.rb:1102:in `rescue in merge'
	/home/kostya/.rbenv/versions/2.3.3/lib/ruby/2.3.0/uri/generic.rb:1099:in `merge'
	/home/kostya/.rbenv/versions/2.3.3/lib/ruby/2.3.0/webrick/httpresponse.rb:277:in `setup_header'
	/home/kostya/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/rack-1.6.8/lib/rack/handler/webrick.rb:16:in `setup_header'
	/home/kostya/.rbenv/versions/2.3.3/lib/ruby/2.3.0/webrick/httpresponse.rb:207:in `send_response'
[...]

PR #1091, opened 2018-02-14
confirmation should redirect to default_confirm_success_url by default

maysam:
#1084

### DIFF
--- a/app/controllers/devise_token_auth/confirmations_controller.rb
+++ b/app/controllers/devise_token_auth/confirmations_controller.rb
@@ -20,8 +20,15 @@ module DeviseTokenAuth
         redirect_headers = build_redirect_headers(token,
                                                   client_id,
                                                   redirect_header_options)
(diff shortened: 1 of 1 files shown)


dks17:
It needs some tests covering the case (then `redirect_url` is absent).
Is controller suitable place to put your code? What if I need to get `redirect_url` in other controller of an application?

maysam:
this is already following the pattern in [registrations_controller.rb#L20-L23](https://github.com/maysam/devise_token_auth/blob/12c87ebbcab2f94c7e8fb7198a48cfadd6c8ca40/app/controllers/devise_token_auth/registrations_controller.rb#L20-L23)

dks17:
@maysam you are right.
I think that will be better to put your code and you the pattern in one place.
And anyway tests are needed.
I think.

MaicolBen:
Merging as the author didn't show up anymore and it isn't mandatory in this case the tests as it's pretty simple.

PR #1095, opened 2018-02-20
Return locked out message from sessions controller when resource is locked

TylerRockwell:
### Why?
Currently, when a user account is locked, devise token auth returns an error stating:
```
A confirmation email was sent to your account at user@example.com.
You must follow the instructions in the email before your account can be activated
```
This message can be misleading, especially when the `:confirmable` module is not in use.

### What Changed?
When an account is locked devise token auth now returns this message:
```
Your account has been locked due to an excessive number of unsuccessful sign in attempts.
```


zachfeldman:
Makes sense to me! If another contributor approves, I'll merge.

TylerRockwell:
Excellent. Thanks for the quick response and merge @zachfeldman and @MaicolBen .

zachfeldman:
Thanks for your contribution!

PR #1096, opened 2018-02-20
Message for the Lockable module

boris-korkmazov:


zachfeldman:
@boris-korkmazov this looks pretty similar to another PR just submitted?

PR #1098, opened 2018-02-21
Further refactor of token creation

Evan-M:
I commented on #1061 after it was merged, but since I'm not sure I was very clear, I went ahead and further refactored things to DRY up the code a bit more.

Essentially, the [`#create_token`] method can accept arbitrary additional keyword attributes that will be merged into the resulting `tokens[client_id]` hash.  This allows [`#create_new_auth_token` ]to reuse the token creation logic, instead of duplicating it. [See: `e601951`][`e601951`] for details.

Note the first commit of this PR is misc cleanup, and perhaps doesn't belong.

I didn't add additional tests, since there is no additional functionality.

[`#create_token`]: https://github.com/lynndylanhurley/devise_token_auth/blob/e601951c10ae57053b33e4090f7404115df4d1e4/app/models/devise_token_auth/concerns/user.rb#L91

[`#create_new_auth_token`]: https://github.com/lynndylanhurley/devise_token_auth/blob/e601951c10ae57053b33e4090f7404115df4d1e4/app/models/devise_token_auth/concerns/user.rb#L170

[`e601951`]: https://github.com/lynndylanhurley/devise_token_auth/commit/e601951c10ae57053b33e4090f7404115df4d1e4

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -42,17 +42,9 @@ module DeviseTokenAuth::Concerns::User
     before_save :remove_tokens_after_password_reset
 
     # don't use default devise email validation
-    def email_required?
-      false
-    end
-
-    def email_changed?
-      false
-    end
-
-    def will_save_change_to_email?
-      false
-    end
+    def email_required?; false; end
+    def email_changed?; false; end
+    def will_save_change_to_email?; false; end
 
     def password_required?
       return false unless provider == 'email'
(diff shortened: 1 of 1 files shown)


zachfeldman:
Hey @Evan-M thanks for this! Your refactorings make sense and I get what you're doing here. If one other person approves this PR I'll merge it.

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -42,87 +42,68 @@ def self.tokens_match?(token_hash, token)
...
-    def email_changed?
-      false
-    end
-
-    def will_save_change_to_email?
-      false
-    end
+    def email_required?; false; end
+    def email_changed?; false; end
+    def will_save_change_to_email?; false; end
```
KelseyDH: These unexplained `devise` overrides are the bane of my existence and the cause of a lot of hair pulling for a lot of developers...  But alas, you aren't changing their functionality here.  

That said, I'd just leave these alone, so we can preserve our git blame history on these three methods to help the next poor soul who encounters this gotcha while wondering what's magically overriding their email callback on their user object.

KelseyDH:
Could there be any reason for wanting to preserve our calls to `self.foo` on this concern which initially was designed to target the user model?

If not, then this looks like a great a refactor. 👍

app/models/devise_token_auth/concerns/user.rb
zachfeldman: @KelseyDH is there a pr to fix the devise overrides situation? If not, feel free to open  so we can debate!

PR #1109, opened 2018-03-12
max_number_of_devices should be used in a new session as well

### LINKED ISSUE #637: make use of max_number_of_devices on sign_in
Hi everybody, 

The `create_new_auth_token` method in `DeviseTokenAuth::Concerns::User` actually maintains the maximum count of tokens per user according to the `max_number_of_devices` parameter. 

But create method of `DeviseTokenAuth::SessionsController` does not use `create_new_auth_token` so a new token is added on each login. 

Is this something that needs to be fixed?

Thanks, Hasan.

### LINKED ISSUE #1107: max_number_of_devices config seems doesn't work!
I use devise_token_auth as the authentication for my rails application alongside angular2-token. when i set max_number_of_devices to 1 in the config file, it seems doesn't work for me. I can still log in to my app with more than 1 devise.

this is my Gemfile configurations
```
# Authentication
gem 'devise'
gem 'devise_token_auth', :github => 'lynndylanhurley/devise_token_auth', :branch => 'master'
gem 'rack-cors', :require => 'rack/cors'
```
this is my devise_token_auth config file
```
  # By default the authorization headers will change after each request. The
  # client is responsible for keeping track of the changing tokens. Change
  # this to false to prevent the Authorization header from changing after
  # each request.
  config.change_headers_on_each_request = false

  # By default, users will need to re-authenticate after 2 weeks. This setting
  # determines how long tokens will remain valid after they are issued.
  config.token_lifespan = 24.hours

  # Sets the max number of concurrent devices per user, which is 10 by default.
  # After this limit is reached, the oldest tokens will be removed.
  config.max_number_of_devices = 1

  # Sometimes it's necessary to make several requests to the API at the same
  # time. In this case, each request in the batch will need to share the same
  # auth token. This setting determines how far apart the requests can be while
  # still using the same auth token.
  # config.batch_request_buffer_throttle = 5.seconds

  # This route will be the prefix for all oauth2 redirect callbacks. For
  # example, using the default '/omniauth', the github oauth2 provider will
  # redirect successful authentications to '/omniauth/github/callback'
  # config.omniauth_prefix = "/omniauth"

  # By default sending current password is not needed for the password update.
  # Uncomment to enforce current_password param to be checked before all
  # attribute updates. Set it to :password if you want it to be checked only if
  # password is updated.
  config.check_current_password_before_update = :attributes

  # By default we will use callbacks for single omniauth.
  # It depends on fields like email, provider and uid.
  # config.default_callbacks = true

  # Makes it possible to change the headers names
  # config.headers_names = {:'access-token' => 'access-token',
  #                        :'client' => 'client',
  #                        :'expiry' => 'expiry',
  #                        :'uid' => 'uid',
  #                        :'token-type' => 'token-type' }

  # By default, only Bearer Token authentication is implemented out of the box.
  # If, however, you wish to integrate with legacy Devise authentication, you can
  # do so by enabling this flag. NOTE: This feature is highly experimental!
  # config.enable_standard_devise_support = false
end

```

Anyone can help me... Thanks

MaicolBen:
Resolves #1107
Resolves #637

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -98,6 +98,8 @@ module DeviseTokenAuth::Concerns::User
       expiry: expiry
     }.merge!(token_extras)
 
+    clean_old_tokens
+
     [client_id, token, expiry]
   end
 
@@ -196,16 +198,11 @@ module DeviseTokenAuth::Concerns::User
 
   def update_auth_header(token, client_id='default')
     headers = build_auth_header(token, client_id)
-    while tokens.length > 0 && DeviseTokenAuth.max_number_of_devices < tokens.length
-      oldest_client_id, _tk = tokens.min_by { |_cid, v| v[:expiry] || v["expiry"] }
-      tokens.delete(oldest_client_id)
-    end
-
+    clean_old_tokens
     save!
 
     headers
   end
-
 
   def build_auth_url(base_url, args)
     args[:uid]    = uid
@@ -213,7 +210,6 @@ module DeviseTokenAuth::Concerns::User
 
     DeviseTokenAuth::Url.generate(base_url, args)
   end
-
 
   def extend_batch_buffer(token, client_id)
     self.tokens[client_id]['updated_at'] = Time.now
@@ -257,4 +253,10 @@ module DeviseTokenAuth::Concerns::User
     end
   end
 
+  def clean_old_tokens
+    while tokens.length > 0 && DeviseTokenAuth.max_number_of_devices < tokens.length
+      oldest_client_id, _tk = tokens.min_by { |_cid, v| v[:expiry] || v["expiry"] }
+      tokens.delete(oldest_client_id)
+    end
+  end
 end
--- a/test/controllers/devise_token_auth/sessions_controller_test.rb
+++ b/test/controllers/devise_token_auth/sessions_controller_test.rb
@@ -72,6 +72,33 @@ class DeviseTokenAuth::SessionsControllerTest < ActionController::TestCase
             assert_equal '0.0.0.0', @new_last_sign_in_ip
           end
         end
+
+        describe "with multiple clients and headers don't change in each request" do
+          before do
+            DeviseTokenAuth.max_number_of_devices = 1
+            DeviseTokenAuth.change_headers_on_each_request = false
+            @tokens = []
+            (1..3).each do |n|
+              post :create,
+                   params: {
+                     email: @existing_user.email,
+                     password: 'secret123'
+                   }
+              @tokens << @existing_user.reload.tokens
+            end
+          end
+
+          test 'should delete old tokens' do
+            current_tokens = @existing_user.reload.tokens
+            assert_equal 1, current_tokens.count
+            assert_equal @tokens.pop.keys.first, current_tokens.keys.first
+          end
+
+          after do
+            DeviseTokenAuth.max_number_of_devices = 10
+            DeviseTokenAuth.change_headers_on_each_request = true
+          end
+        end
       end
 
       describe 'get sign_in is not supported' do


MaicolBen:
Wait a minute, I removed the code and the test is still working...

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -257,4 +255,10 @@ def remove_tokens_after_password_reset
     end
   end
 
+  def clean_old_tokens
```
Evan-M: Isn't this code technically <code>O(n<sup>2</sup>)</code>? (The `Enumberable#min_by` call is inside a `while` loop).

Even though `Enumberable#min_by` returns a single key/value pair, it still iterates over the entire collection every time it's called.

I would think that it is more efficient to iterate over the collection once, sorting the key/value pairs by `:expiry` of the value, and then shift the oldest key/values from tokens in a while loop.

```ruby
  ...
    protected
    ...
    def max_client_tokens_exceeded?
      tokens.length > DeviseTokenAuth.max_number_of_devices
    end

    def clean_old_tokens
      if tokens.present? && max_client_tokens_exceeded?
        # Using Enumerable#sort_by on a Hash will typecast it into an associative
        #   Array (i.e. an Array of key-value Array pairs). However, since Hashes
        #   have an internal order in Ruby 1.9+, the resulting sorted associative
        #   Array can be converted back into a Hash, while maintaining the sorted
        #   order.
        self.tokens = tokens.sort_by { |_cid, v| v[:expiry] || v['expiry'] }.to_h

        # Since the tokens are sorted by expiry, shift the oldest client token
        #   off the Hash until it no longer exceeds the maximum number of clients
        tokens.shift while max_client_tokens_exceeded?
      end
    end
  ...
```

test/controllers/devise_token_auth/sessions_controller_test.rb
```diff
@@ -72,6 +72,23 @@ class DeviseTokenAuth::SessionsControllerTest < ActionController::TestCase
             assert_equal '0.0.0.0', @new_last_sign_in_ip
           end
         end
+
```
Evan-M: Shouldn't you assert that the _correct_ tokens (i.e. the _oldest_ tokens) are deleted here?

E.g.: 
 Given that the `max_number_of_devices = 5`, and there are 5 existing tokens,
 When an additional token is added, the oldest token is dropped and the new one is present.

MaicolBen:
The code was correct but the test doesn't, this happens when `change_headers_on_each_request` is false, so I missed that in the test

app/models/devise_token_auth/concerns/user.rb
MaicolBen: This change breaks randomically the test that you mentioned: `test/controllers/demo_user_controller_test.rb#L506`. I tried for a day fixing it but it's quite weird (and hard to debug with minitest). If we merge the PR without this change, can you create a new one and try to fix it with your comment?

Evan-M: Yeah, no problem. I can submit a second PR. Although, I think I saw random test failures on travis-ci before I even made my comment?

MaicolBen: Yes, but they weren't random, they were my errors, I haven't updated the PR yet

Evan-M: Is the debugging a challenge because the `Minitest::Reporters::ProgressReporter` leaves out line numbers? I've definitely been challenged with debugging failing tests while using that reporter.  While I prefer the `ProgressReporter`, I find that the default minitest reporter provides much more useful debugging info.

I usually just temporarily comment out [line 27 of the `test_helper.rb`](https://github.com/lynndylanhurley/devise_token_auth/blob/master/test/test_helper.rb#L27) while running these tests. YMMV.

I've been meaning to look into the the issue and submit a PR to [`kern/minitest-reporters`](https://github.com/kern/minitest-reporters) with a fix.

MaicolBen: Thanks for the tip! It doesn't throw a backtrace either 😞

MaicolBen:
Fixed test, @Evan-M will be making another PR improving `clean_old_tokens`, thank you for your comments!

Evan-M:
@MaicolBen: I checked out the [`max_number_of_devices_in_new_session`](https://github.com/maicolben/devise_token_auth/tree/max_number_of_devices_in_new_session) branch from your forked repo and added my code on top of your PR. So lynndylanhurley/devise_token_auth#1115 I think includes this PR as well.

I think when this PR (#1109) is merged, mine will update since all the commit hashes should match up. If not, I can re-issue my PR against `master` once #1109 is merged.

MaicolBen:
@Evan-M You can merge this PR as it has Zach's approve, otherwise we can close this one in favor of #1115

Evan-M:
@MaicolBen I've only got `Contributor` permissions (not `Collaborator`) so unfortunately I cannot merge PR.

MaicolBen:
I didn't know, sorry, can you approve?

PR #1113, opened 2018-03-18
Refactor uses of time now

### LINKED ISSUE #1112: Tests on token expiry fail when they're run on the WET time zone
While working on a different issue, I found one of the tests on token expiry (`test/models/user_test.rb`, on the test `'works per user'`) failed for no apparent reason.

What I found was that, although `DeviseTokenAuth.token_lifespan` is set to 2.weeks, the expression `(Time.now + DeviseTokenAuth.token_lifespan).to_i - Time.now.to_i` returns 13 days, 23 hours and 59 minutes. When changing the system's time zone to e.g. EST (the local one is WET), this no longer happens.

The fix is rather simple, one just needs to replace all instances of `Time.now` to `Time.zone.now` or `Time.current`. I'll open the pull request shortly.

nesteves:
Fixes #1112 by replacing all instances of `Time.now` with `Time.zone.now`.

I also removed a duplicate test from `test/controllers/devise_token_auth/confirmations_controller_test.rb` that included a block that was never run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,7 +500,7 @@ Models that include the `DeviseTokenAuth::Concerns::User` concern will have acce
   # store client + token in user's token hash
   @resource.tokens[client_id] = {
     token: BCrypt::Password.create(token),
-    expiry: (Time.now + @resource.token_lifespan).to_i
+    expiry: (Time.zone.now + @resource.token_lifespan).to_i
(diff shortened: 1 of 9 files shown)


zachfeldman:
All the tests pass.

PR #1115, opened 2018-03-19
Max number of devices in new session

Evan-M:
Additional refactoring, per discussion in lynndylanhurley/devise_token_auth#1109.

See: https://github.com/lynndylanhurley/devise_token_auth/pull/1109#pullrequestreview-103226402 and https://github.com/lynndylanhurley/devise_token_auth/pull/1109#issuecomment-373167880

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -60,7 +60,9 @@ module DeviseTokenAuth::Concerns::SetUserByToken
       if devise_warden_user && devise_warden_user.tokens[@client_id].nil?
         @used_auth_by_token = false
         @resource = devise_warden_user
-        @resource.create_new_auth_token
+        # REVIEW: The following line _should_ be safe to remove;
+        #  the generated token does not get used anywhere.
+        # @resource.create_new_auth_token
       end
     end
 
--- a/test/controllers/demo_user_controller_test.rb
+++ b/test/controllers/demo_user_controller_test.rb
@@ -407,6 +407,53 @@ class DemoUserControllerTest < ActionDispatch::IntegrationTest
           DeviseTokenAuth.headers_names[:'access-token'] = 'access-token'
         end
       end
+
+      describe 'maximum concurrent devices per user' do
+        before do
+          # Set the max_number_of_devices to a lower number
+          #  to expedite tests! (Default is 10)
+          DeviseTokenAuth.max_number_of_devices = 5
+
+          # @max_devices = DeviseTokenAuth.max_number_of_devices
+        end
+
+        it 'should limit the maximum number of concurrent devices' do
+          # increment the number of devices until the maximum is exceeded
+          1.upto(DeviseTokenAuth.max_number_of_devices + 1).each do |n|
+            # initial_tokens = @resource.reload.tokens
+
+            assert_equal(
+              [n, DeviseTokenAuth.max_number_of_devices].min,
+              @resource.reload.tokens.length
+            )
+
+            # Add a new device (and token) ahead of the next iteration
+            @resource.create_new_auth_token
+
+            # refute_equal initial_tokens, @resource.reload.tokens
+          end
+        end
+
+        it 'should drop the oldest token when the maximum number of devices is exceeded' do
+          # create the maximum number of tokens
+          1.upto(DeviseTokenAuth.max_number_of_devices).each do
+            @resource.create_new_auth_token
+          end
+
+          # get the oldest token
+          oldest_token, _ = @resource.reload.tokens \
+                              .min_by { |cid, v| v[:expiry] || v["expiry"] }
+
+          # create another token, thereby dropping the oldest token
+          @resource.create_new_auth_token
+
+          assert_not_includes @resource.reload.tokens.keys, oldest_token
+        end
+
+        after do
+          DeviseTokenAuth.max_number_of_devices = 10
+        end
+      end
     end
 
     describe 'bypass_sign_in' do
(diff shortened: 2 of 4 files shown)


MaicolBen:
@Evan-M You need to rebase

Evan-M:
@MaicolBen rebased.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -60,7 +60,9 @@ def set_user_by_token(mapping=nil)
       if devise_warden_user && devise_warden_user.tokens[@client_id].nil?
         @used_auth_by_token = false
         @resource = devise_warden_user
-        @resource.create_new_auth_token
+        # REVIEW: The following line _should_ be safe to remove;
+        #  the generated token does not get used anywhere.
+        # @resource.create_new_auth_token
```
MaicolBen: `update_auth_header` should use this, but if you say so, we can remove it

test/controllers/demo_user_controller_test.rb
```diff
@@ -407,6 +407,53 @@ class DemoUserControllerTest < ActionDispatch::IntegrationTest
...
         end
       end
+
+      describe 'maximum concurrent devices per user' do
+        before do
+          # Set the max_number_of_devices to a lower number
+          #  to expedite tests! (Default is 10)
+          DeviseTokenAuth.max_number_of_devices = 5
+
+          # @max_devices = DeviseTokenAuth.max_number_of_devices
```
MaicolBen: remove the commented line, or is this documentation?

test/controllers/demo_user_controller_test.rb
```diff
@@ -407,6 +407,53 @@ class DemoUserControllerTest < ActionDispatch::IntegrationTest
...
+
+            assert_equal(
+              [n, DeviseTokenAuth.max_number_of_devices].min,
+              @resource.reload.tokens.length
+            )
+
+            # Add a new device (and token) ahead of the next iteration
+            @resource.create_new_auth_token
+
+            # refute_equal initial_tokens, @resource.reload.tokens
```
MaicolBen: why this refute is commented?

test/controllers/devise_token_auth/sessions_controller_test.rb
```diff
@@ -75,23 +75,46 @@ class DeviseTokenAuth::SessionsControllerTest < ActionController::TestCase
...
-          test 'should delete old tokens' do
-            current_tokens = @existing_user.reload.tokens
-            assert_equal 1, current_tokens.count
-            assert_equal @tokens.pop.keys.first, current_tokens.keys.first
+          test 'should drop old tokens when max number of devices is exceeded' do
+            1.upto(DeviseTokenAuth.max_number_of_devices).each do |n|
+              post :create, params: @user_session_params
+            end
+
+            oldest_token, _ = @existing_user.reload.tokens \
```
MaicolBen: no need for `, _`

Evan-M: That is purely convention, although I used it incorrectly...

https://github.com/bbatsov/ruby-style-guide#trailing-underscore-variables

It _should_ either be:
~~~ruby
oldest_token, = \
  @existing_user.reload.tokens \
    .min_by { |cid, v| v[:expiry] || v["expiry"] } # => [ <"client_id">, {...<token>...} ]
~~~
or
~~~ruby
oldest_client_id, _oldest_token = \ 
  @existing_user.reload.tokens \
    .min_by { |cid, v| v[:expiry] || v["expiry"] } # => [ <"client_id">, {...<token>...} ]
~~~

If you _really_ truly think I should remove the `, _` entirely, I will add a comment that an array is returned by `#min_by`.

On a different note, I've been itching to create an issue to get rubocop setup on this repo, so that stylistic issues like this are a non-issue.

test/controllers/demo_user_controller_test.rb
Evan-M: It is cruft and I didn't mean to leave it in the commit! 😆

test/controllers/devise_token_auth/sessions_controller_test.rb
MaicolBen: You're right, thank you for the long explanation, and if you enable rubocop, there will be tons of issues, that's why we have Code Climate that analyzes only the differences. I will create an issue for it 😄

test/controllers/demo_user_controller_test.rb
Evan-M: I actually meant to remove it.  While the test still passes with it uncommented (along with line 423 above), it really doesn't add anything to the test here.  We've already asserted that the number of concurrent tokens/devices is correct; if anything the addition of this test will make the test more brittle.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
Evan-M: Rather confusingly, there are _two_ methods with that name. 
the model concern:
https://github.com/lynndylanhurley/devise_token_auth/blob/38d27cfa80042c184c6f69e74c0789ca8ec0e407/app/models/devise_token_auth/concerns/user.rb#L199-L205
and the controller concern:
https://github.com/lynndylanhurley/devise_token_auth/blob/38d27cfa80042c184c6f69e74c0789ca8ec0e407/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L96-L154

I'll assume you mean the one in the controller concern (the same file as this change)...

Regardless, I'm not sure I understand what you mean. This is inside a conditional block for the use case of when `DeviseTokenAuth.enable_standard_devise_support = true` and the user has already authenticated with Devise  in the _session_ via Warden. There is no need for an auth token in that case, so why would we create one here?

zachfeldman:
Tests pass. Merging away!

PR #1132, opened 2018-04-04
1131 fix deprecation warning for dirty attributes for rails > 5

### LINKED ISSUE #1131: Clear ActiveRecord::AttributeMethods::Dirty Deprecated Methods
`devise_token_auth/concerns/user.rb:247` is using deprecated methods, littering logs with depracation warning messages.`

```
devise_token_auth
  revision: d4c0e91ccba8b6e6d611fe828e1def442199ac29
  branch: master
  specs:
    devise_token_auth (0.1.43)
      devise (> 3.5.2, < 4.5)
      rails (< 6)

rails (= 5.1.4)
```
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from remove_tokens_after_password_reset at /Users/marin/.rvm/gems/ruby-2.5.0@livestorm-2.5/bundler/gems/devise_token_auth-d4c0e91ccba8/app/models/devise_token_auth/concerns/user.rb:247)
```

Marinlemaignan:
Resolves #1131

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -243,9 +243,18 @@ module DeviseTokenAuth::Concerns::User
     end
   end
 
+  def should_remove_tokens_after_password_reset?
+    if Rails::VERSION::MAJOR <= 5
+      encrypted_password_changed? &&
+      DeviseTokenAuth.remove_tokens_after_password_reset
+    else
+      saved_change_to_encrypted_password? &&
+      DeviseTokenAuth.remove_tokens_after_password_reset
+    end
+  end
+
   def remove_tokens_after_password_reset
-    return unless encrypted_password_changed? &&
-                DeviseTokenAuth.remove_tokens_after_password_reset
+    return unless should_remove_tokens_after_password_reset?
 
     if tokens.present? && tokens.many?
       client_id, token_data = tokens.max_by { |cid, v| v[:expiry] || v["expiry"] }


app/models/devise_token_auth/concerns/user.rb
```diff
@@ -243,9 +243,18 @@ def destroy_expired_tokens
     end
   end
 
+  def should_remove_tokens_after_password_reset?
+    if Rails::VERSION::MAJOR <= 5
+      encrypted_password_changed? &&
+      DeviseTokenAuth.remove_tokens_after_password_reset
+    else
+      saved_change_to_encrypted_password? &&
```
MaicolBen: Seems coherent with https://github.com/plataformatec/devise/blob/dac90eeb80ac351697e0a8d1bda9d1697c7b7078/lib/devise/models/database_authenticatable.rb#L182

zachfeldman:
Thanks for the contribution @Marinlemaignan !

dks17:
```ruby
    if Rails::VERSION::MAJOR <= 5
      encrypted_password_changed? &&
      DeviseTokenAuth.remove_tokens_after_password_reset
    else
      saved_change_to_encrypted_password? &&
      DeviseTokenAuth.remove_tokens_after_password_reset
    end
```

It looks like in the else block is dead code because of expression `Rails::VERSION::MAJOR <= 5` always return true for supported version `4` and `5`? Else block is executed when major rails version is `6` or higher.

@Marinlemaignan in the header of the issue you used strict comparison `> 5`. Did you test it against rails `6`?

@MaicolBen what do you think?

MaicolBen:
Yeah, I would change it to `Devise.activerecord51?` like the code linked by me, good discovery!

dks17:
1. Method `saved_change_to_attribute?` was introduced in `ActiveRecord` since `5.1` version.
2. Version checking should be put outside of the `should_remove_tokens_after_password_reset?` method. Like the code you linked above.
3. `Mongoid` doesn't support `saved_change_to_attribute?` method. I think be better define the `should_remove_tokens_after_password_reset?` method into `mongoid_support` and `active_record_support` concerns. In `active_record_support` concern you may use`ActiveRecord` constant for version checking (didn't check) or `Devise` constant.

PR #1139, opened 2018-04-09
Fix broken stubbed method expectation

### LINKED ISSUE #1137: Breaking tests / travis builds - Mocha gem was updated.
A recent update to the Mocha gem ([release 1.5.0](https://github.com/freerange/mocha/blob/master/RELEASE.md#150)) prevents the use of Mocha outside of test/examples.

This is breaking all of the Travis CI builds, since they try to bundle the latest release.

The simplest solution is probably to specify `~> 1.4.0` in the `Gemfile`. 

That said, it seems like the only place Mocha is used in this project is in `/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb`.  Perhaps the better long term approach is to refactor that test to either: (A) not stub method calls in the `before` blocks,  (B) stub the methods with Minitest instead (thereby removing the dependency on Mocha), or (C) replace the stubbed responses with real data and not stub any methods at all (thereby removing the dependency on Mocha).

Evan-M:
- Update file to require correct Mocha lib for MiniTest (instead of
   Test::Unit). Using the wrong lib previously, meant that Mocha was
   never verifying the stubbed method, and the test passed
   (erroneously).

 - Update test to expect the correct method call for sign_in.

Resolves #1137 (which was exposed by the merge of #1134).

### DIFF
--- a/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb
+++ b/test/controllers/devise_token_auth/omniauth_callbacks_controller_test.rb
@@ -1,5 +1,5 @@
 require 'test_helper'
-require 'mocha/test_unit'
+require 'mocha/minitest'
 
 #  was the web request successful?
 #  was the user redirected to the right page?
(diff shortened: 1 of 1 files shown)


Evan-M:
@zachfeldman, @MaicolBen, @lynndylanhurley, @krzysiek1507 (or anyone else):

Could a few people review this PR?  It resolves the broken builds on [**master**](https://github.com/lynndylanhurley/devise_token_auth/tree/master) following the merge of #1134.

Evan-M:
As soon as I get a second approval, I will merge this!

PR #1142, opened 2018-04-11
Setup appraisal for rails 4

krzysiek1507:


Evan-M:
This is probably a completely separate issue from this PR, but Ruby 2.2.10 was the last expected release of the 2.2 series.  The EOL date for Ruby 2.2 was scheduled for 3/31/2018 (2.5 weeks ago).  Removing `2.2.8` from the `.travis.yml` file will cut out 25% of the current CI builds. (Each appraisal accounts for 9 builds atm)

I guess the bigger question is: how long should this gem support Ruby 2.2, given that it is EOL?

MaicolBen:
@Evan-M [2.1 and 2.2 are still widely used](https://semaphoreci.com/blog/2017/11/08/ruby-versions-used-in-commercial-projects-in-2017.html). I think we can keep supporting ruby 2.2 until we bump into a problematic PR that cannot be fixed without removing it (or a ruby method that increases the performance)

Issue #1143, opened 2018-04-12
Extract Registrations Controller logic out to overrideable methods ???


Marinlemaignan:
Hello there,
I'm facing issues in my code right now because of some lack of flexibility in some methods,  for exemple, `#create` in `RegistrationsController` (i am implementing some rather complicated login logic with scopes all the way through some STI)
My idea would be to extract the resource's initialization code out to an overrideable method
(a bit like when you had moved out those methds to the ResourceFinder Concern which super super useful !)

```
      @resource = resource_class.new(sign_up_params.except(:confirm_success_url))
      @resource.provider   = provider

      # honor devise configuration for case_insensitive_keys
      if resource_class.case_insensitive_keys.include?(:email)
        @resource.email = sign_up_params[:email].try :downcase
      else
        @resource.email = sign_up_params[:email]
      end
```

should become something like :
```
build_registration_resource
```

which would then set the resource:
```
def build_registration_resource
    @resource = resource_class.new(sign_up_params.except(:confirm_success_url))
      @resource.provider   = provider

      # honor devise configuration for case_insensitive_keys
      if resource_class.case_insensitive_keys.include?(:email)
        @resource.email = sign_up_params[:email].try :downcase
      else
        @resource.email = sign_up_params[:email]
      end
end
```

What do you think about that, could that be something you'd be up for? or just a bad idea?

Marinlemaignan:
here's the PR : https://github.com/lynndylanhurley/devise_token_auth/pull/1144

PR #1144, opened 2018-04-12
1143 extract @resource initialization out to a named overrideable method

Marinlemaignan:
https://github.com/lynndylanhurley/devise_token_auth/issues/1143


app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -3,19 +3,11 @@ class RegistrationsController < DeviseTokenAuth::ApplicationController
     before_action :set_user_by_token, only: [:destroy, :update]
     before_action :validate_sign_up_params, only: :create
     before_action :validate_account_update_params, only: :update
+    before_action :build_resource_for_registration, only: [:create]
```
MaicolBen: I believe that would be clearer if this is part of the method instead of a before filter, so right after `def create`, call `build_resource_for_registration`

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -105,6 +95,18 @@ def account_update_params
 
     protected
 
+    def build_resource
```
Evan-M: Seeing the `@resource` instance variable over and over again, makes me wonder if this business logic for building a resource belongs inside the model (or model concern) instead?

For example:

~~~ruby
# In the resource_class model (via ActiveSupport::Concern, or otherwise):
module ClassMethods
  def new_registration(params)
    params[:email].try(:downcase!) if case_insensitive_keys.include?(:email)
    new(params.except(:confirm_success_url))
  end
end
~~~

~~~ruby
# In DeviseTokenAuth::RegistrationsController#create:
# ...

def create
  @resource = resource_class.new_registration(sign_up_params)

  # ...
end

# ...
~~~

You could still override the extracted method to customize the behavior, just inside your model instead of the controller.

Evan-M: On second thought, given how I refactored things in my example above, I think a bulk of the existing logic involves manipulating parameters, and most of it could live inside the `#sign_up_params` protected method that handles the strong_parameters.  That would leave just a plain `@resource = resource_class.new(sign_up_params)` call in the `#create` action.

... Of course, that would not leave a method for you to override anymore.

Another approach might be to extract that `@resource = resource_class.new(sign_up_params)` into a `before_action` callback.  Then you could maintain the ability to customize the behavior by overriding the callback method.

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -6,16 +6,8 @@ class RegistrationsController < DeviseTokenAuth::ApplicationController
...
-      @resource.provider   = provider
-
-      # honor devise configuration for case_insensitive_keys
-      if resource_class.case_insensitive_keys.include?(:email)
-        @resource.email = sign_up_params[:email].try :downcase
-      else
-        @resource.email = sign_up_params[:email]
-      end
-
+      build_resource
```
Evan-M: Consider removing this explicit call to `build_resource` from the `#create` action, and instead call it with `before_action :build_resource, only: [:create]`.

Marinlemaignan: @Evan-M we discussed that with @MaicolBen and i had it the 'callback way' before, but then we moved it to an explicit call, which kind of makes sense when looking at [devise](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/registrations_controller.rb#L17)

app/controllers/devise_token_auth/registrations_controller.rb
Marinlemaignan: - about the concern: yes my first idea was to have something like the [ResourceFinder](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/resource_finder.rb#L20)Concern. But for building. And i'd prefer have it stay in the Controllers rather than the models.

app/controllers/devise_token_auth/registrations_controller.rb
Evan-M: @Marinlemaignan Oops, I _should_ have been more diligent in reading through the discussion in #1143. Carry on! 🎒

app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -88,22 +80,33 @@ def destroy
...
-      params.permit([*params_for_resource(:sign_up), :confirm_success_url])
+      params.permit(*params_for_resource(:sign_up))
     end
 
     def account_update_params
       params.permit(*params_for_resource(:account_update))
     end
 
-    protected
+  protected
```
MaicolBen: align with `def`

test/test_helper.rb
```diff
@@ -100,3 +100,10 @@ class TestCase
     include Rails::Controller::Testing::Integration
   end
 end
+
+# ensure any mocha version is correctly required
+begin
+  require 'mocha/minitest'
```
MaicolBen: why this?

Marinlemaignan: https://github.com/freerange/mocha/issues/324
i happened to personnaly ran into this...
could be removed obvs

app/controllers/devise_token_auth/registrations_controller.rb
Marinlemaignan: this is imho a good way to improve scanning a file.
rubocop would allow you to indent in such a way, Rails favors another but very similar one (but i won't tab twice lol), is this debatable ?
otherwise i will do
(rails exemple : https://github.com/rails/rails/blob/d0d7ce001bdbb0e5551da881caf8156c06238b84/actioncable/lib/action_cable/remote_connections.rb#L56)

MaicolBen: your way could be better, but we should stick to the project standard, so maybe you can change all the protected/private alignments in another PR

Marinlemaignan:
@MaicolBen if you could just squash-merge them from here please i'd appreciate (i do love that new option, so useful)

MaicolBen:
@Marinlemaignan sure, @Evan-M can you approve this PR?

test/test_helper.rb
Marinlemaignan: actually this is a little bothering me.. i think it could be quite bothering for the gem's users if something goes really wrong cause i'm not doing anything with the error..
i'll remove that, and open an issue / add to the docs so it's referenced somewhere

MaicolBen: I think if you remove this, it won't break travis. So remove it if you can

Marinlemaignan: yes i'll open an issue for referencing

PR #1154, opened 2018-04-25
Return Auth Headers in Register for Valid Unconfirmed User

### LINKED ISSUE #89: allow_unconfirmed_access_for
Hi,

How do you configure allow_unconfirmed_access_for. I've added 

``` ruby
Devise.setup do |config|   
  config.allow_unconfirmed_access_for = 7.days
end
```

to config/initializers/devise.rb and confirmed with byebug that the configuration is hit, but the users still aren't allowed unconfirmed access.

yonahw:
fixes #89 

Although new valid unconfirmed users are able to login the auth headers are not returned on register.


app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -46,7 +46,9 @@ def create
               client_config: params[:config_name],
               redirect_url: @redirect_url
             })
-          else
+          end
+
+          if !@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?
```
MaicolBen: Move `!@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?` to a method here like `active_for_authentication?` because this controller's method is huge

test/controllers/devise_token_auth/registrations_controller_test.rb
```diff
@@ -81,6 +81,33 @@ class DeviseTokenAuth::RegistrationsControllerTest < ActionDispatch::Integration
       end
     end
 
+    describe 'using allow_unconfirmed_access_for' do
+      before do
+        @original_duration = Devise.allow_unconfirmed_access_for
+        Devise.allow_unconfirmed_access_for = nil
```
MaicolBen: if this is null, how can it return the headers?

yonahw: Setting this to nil means unlimited time period for accessing without confirmation.

See [Devise Docs](http://www.rubydoc.info/github/plataformatec/devise/master/Devise%2FModels%2FConfirmable:confirmation_period_valid%3F)

MaicolBen: That's the solution that posted [Zach in the issue](https://github.com/lynndylanhurley/devise_token_auth/issues/89#issuecomment-333309952). In fact, I cloned your branch, commented `active_for_authentication?` and the tests still pass, so I don't see any improvement or your tests are wrong.

yonahw: Commenting out the active_for_authentication? is not the same as reverting the code change since there was previously an if \ else which is now two separate ifs.

If you are saying that there is no test to make sure that we don't send when the headers when the user is not confirmed I can write a test for that as well. I assumed that there was an existing test covering that case but I suppose that is not the case.

I discovered this issue because I was not receiving the headers on the successful registration.

What Zach posted in the issue is correct, the registration succeeds and the user can subsequently log in. However the auth headers are definitely currently not returned from the registration.

yonahw: Thank you for approving.

MaicolBen:
We can skip this CC issue, I will fix CC configuration once rubocop is added here

MaicolBen:
@yonahw sorry for the delay, seems that I waited for more reviews, can you rebase?

bananatron:
Thanks for this one! Was making a PR and discovered you caught it already.

MaicolBen:
@bananatron do you agree with the change? can you rebase? or you can create a new one with this code

bananatron:
@MaicolBen I do agree with these changes - I've been using essentially the exact same change in production w/ our app for a while now.

I've created https://github.com/lynndylanhurley/devise_token_auth/pull/1238 if we can't manage to get this merged in. 🙏

yonahw:
@MaicolBen sorry for the delay, had some other pressing issues to deal with.

I have rebased.

MaicolBen:
Thanks! but I will close this one because the other one has an improvement in the code

PR #1125, opened 2018-03-24
Add rubocop

Evan-M:
Taking a stab at adding `rubocop` to the repo, as suggested in #1124 

I've already addressed some of the simpler Cops that can utilize the `--auto-correct` option.

Each correction for a given cop was made in a separate commit, so to make reviewing easier (I hope).

The flow I've used was as follows:
  1) Pick a Cop / set of violations in the `.rubocop_todo.yml`
  2) Create a local branch with the Cop name
  3) Make 1 or more updates to code, or move a todo item into the main `.rubocop.yml` config.
  4) Merge the Cop branch back onto the `add_rubocop` branch with the `--no-ff` option.

This PR gets the offenses down to 652, but there is still more work to do!

PR #1156, opened 2018-04-27
Rebase rubocop branch

dks17:
Reference #1125


dks17:
@MaicolBen
How does it look?

.rubocop_todo.yml
```diff
@@ -195,62 +217,46 @@ Lint/UnusedBlockArgument:
...
-    - 'app/controllers/devise_token_auth/registrations_controller.rb'
-    - 'app/controllers/devise_token_auth/sessions_controller.rb'
-    - 'app/controllers/devise_token_auth/unlocks_controller.rb'
-    - 'app/models/devise_token_auth/concerns/user.rb'
-    - 'app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb'
-    - 'app/validators/email_validator.rb'
-    - 'lib/devise_token_auth/engine.rb'
-    - 'lib/devise_token_auth/rails/routes.rb'
-    - 'lib/generators/devise_token_auth/install_generator.rb'
+  Max: 189
```
MaicolBen: why 189?

.rubocop_todo.yml
```diff
@@ -562,3 +549,9 @@ Style/TrailingUnderscoreVariable:
 Style/UnneededPercentQ:
   Exclude:
     - 'app/controllers/devise_token_auth/omniauth_callbacks_controller.rb'
+
+# Offense count: 74
+# Configuration parameters: AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
+# URISchemes: http, https
+Metrics/LineLength:
+  Max: 189
```
MaicolBen: same as above

dks17:
@MaicolBen, as I understand 189 is length of the most longest line in whole project.

Check it with command:
```console
rubocop --only Metrics/LineLength
```
I don't know why exclude list of Metrics/LineLength cop is not shown.

The lines with length more than 80 are persisted in the project and Metrics/LineLength cop was not fixed totally.

@Evan-M What do you think about current PR and why 189?

.rubocop_todo.yml
Evan-M: The `--auto-gen-config` option sets this...  It picks whatever line-length necessary so that everything is ignored. Hence, I set it explicitly back to the default of `Max: 80` in the `.rubocop.yml` file.  I then manually exclude each file with a violation, instead of extending the max length.

MaicolBen: I just realized that this is a todo file, I commented this also because #1125 sets 80. Will you remove this file in the future?

MaicolBen:
I see the purpose of the todo file, the idea is once is everything is fixed, that file is removed, when you rebase again :/, I will merge

Evan-M:
Before merging, I'd like to integrate rubocop with guard, and add some docs.

Also, let's make sure we're methodical in reviewing each commit... There
are a lot, and once we squash them all it'll be harder to review.

Evan-M:
~~Before merging, I'd like to integrate rubocop with guard, and add some docs.~~

~~Also, let's make sure we're methodical in reviewing each commit... There are a lot, and once we squash them all it'll be harder to review.~~

Whoops, I thought you were talking about merging #1125.

UPDATE: I've actually cherry-picked lynndylanhurley/devise_token_auth@a2a67f0 and lynndylanhurley/devise_token_auth@4be88a1 into #1125, and added via `--fixup` commits lynndylanhurley/devise_token_auth@61c582b and lynndylanhurley/devise_token_auth@8f46f25.

So this PR probably doesn't need to be merged.  

See: lynndylanhurley/devise_token_auth@e89be1a and lynndylanhurley/devise_token_auth@c19432b.

MaicolBen:
@Evan-M So can we close this one?

Evan-M:
Closing. See https://github.com/lynndylanhurley/devise_token_auth/pull/1156#issuecomment-385229494.

PR #1169, opened 2018-05-14
Require mocha >= 1.5

### LINKED ISSUE #1149: Mocha/minitest issue in test_helper.rb
Depending on which versions bundler would have installed it can happen to run into [mocha#324](https://github.com/freerange/mocha/issues/324)
we would need to investigate that as to make this library as simple as possible

a quick fix would be 
```
begin
  require 'mocha/minitest
rescue
  require 'mocha/mini_test'
end
```

krzysiek1507:
Require mocha >= 1.5 to fix #1149

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,7 @@ group :development, :test do
   gem 'minitest-focus'
   gem 'minitest-rails'
   gem 'minitest-reporters'
-  gem 'mocha'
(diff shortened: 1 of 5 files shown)


krzysiek1507:
I'll take a look and try to remove unused gems (e.g. figaro, thor).

PR #1184, opened 2018-06-27
Disable email uniqueness validation

ksoderstrom:
This allows a model to disable the default uniqueness validation of email, in order to handle it in a custom way. That's needed if you want to run a multi-tenant setup, which is also how the testing of this change is done.


MaicolBen:
@ksoderstrom Thank you for your work, I agree to separate the uniqueness customization, but I don't agree with the testing approach because devise token auth is agnostic to the database/application design. However, I don't know another approach (is it worth it in this case?), maybe someone else can throw some thoughts.
Also, can you fix travis?

bhushangahire:
How to remove Email uniqueness validation altogether?

PR #1188, opened 2018-07-03
Optimize resource valid check after set the headers

### LINKED ISSUE #1160: If the user is invalid, it doesn't return the tokens in sign in
**Version**: 0.1.43 and redux-token-auth.

I am trying out this gem and while trying out sign_in, I am getting 200 OK result with the user model being responded with in JSON too, but the access-token and client and other such headers are being left out.

What this does in my [Redux-Token-Auth](https://github.com/kylecorbelli/redux-token-auth) is that, they're being saved as `undefined` and therefore there's no persistence at all. Because the verfying token url is thus:
```
http://localhost:3001/auth/validate_token?access-token=undefined&client=undefined&uid=undefined
```

This is a bummer considering this is what the gem is all about.

MaicolBen:
Resolves #1160

In case `change_headers_on_each_request` is disabled: this will improve and prevent no-tokens sent if the user is invalid.

In case `change_headers_on_each_request` is enabled: this will throw an error if the user is invalid even after loading from DB (password controller and other user's controllers needs to send errors to the fronted), if this happens, it means the user left the model in a valid state and it's a bug of the developer's application.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -99,7 +99,8 @@ module DeviseTokenAuth::Concerns::SetUserByToken
 
   def update_auth_header
     # cannot save object if model has invalid params
-    return unless defined?(@resource) && @resource && @resource.valid? && @client_id
+
+    return unless @resource && @client_id
 
     # Generate new client_id with existing authentication
     @client_id = nil unless @used_auth_by_token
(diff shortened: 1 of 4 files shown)


MaicolBen:
@dks17 do you approve this PR? need someone that reviews this

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -115,54 +115,61 @@ def update_auth_header
...
-            auth_header = @resource.create_new_auth_token(@client_id)
-          end
-
-          # update the response header
-          response.headers.merge!(auth_header)
-
-        end # end lock
-      end # end ensure_pristine_resource
+      unless @resource.valid?
+        @resource = resource_class.find(@resource.id) # errors are keep after reload
```
dks17: Why don't you use `find_resource` method from `ResourceFinder`?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -115,54 +115,61 @@ def update_auth_header
...
+        # if we left the model in a bad state, something is wrong in our app
+        raise ActiveRecord::RecordInvalid unless @resource.valid?
```
dks17: `ActiveRecord`'s and `Mongoid`'s `.find` method return exceptions with different names. I think we should avoid such exceptions.

Where will you catch `ActiveRecord::RecordInvalid` error?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -115,54 +115,61 @@ def update_auth_header
...
-          else
-            auth_header = @resource.create_new_auth_token(@client_id)
-          end
-
-          # update the response header
-          response.headers.merge!(auth_header)
-
-        end # end lock
-      end # end ensure_pristine_resource
+      unless @resource.valid?
```
dks17: Does this `while` block actually need?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: As the readme states, this helps the developer fix his applicacion, the user was saved invalid somewhere (outside the gem) and now it doesn't let store the tokens because it's invalid. It happened to me sometimes. We can throw a custom exception that says the model is invalid, wdyt?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: I am not looking for email, just reloading the `@resource`, it should be done this way because errors aren't gone after `@resource.reload`, I know it's kind of nasty solution but I can't find something better, check the comment below in order to see why I'm looking for errors.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: yes, in order to avoid going to the db twice all the time, we need to check if it's really invalid (reloading the resource) if it's invalid

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
dks17: I'm OK with custom exception.

Does front-end client understand that something gone wrong (returned error message)?

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
dks17: `find_resource` is suitable is not only for email. Yes, `find_resource` is unnecessarily complex. I have already fix `find_resource` in `mongoid` pr and the method works with different ORMs.
Use `@resource.to_param` instead of `@resource.id`. https://api.rubyonrails.org/classes/ActiveModel/Conversion.html#method-i-to_param

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: This should lead to 500, it's a bug in the developer's software, the model is invalid, so it isn't expected and should be fixed

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
MaicolBen: doesn't Mongo have id? I will use `to_param`, thanks

dks17: `id` is custom field. `_id` is unique identifier of document in collection and it is storing as BSON. So, you should use `user._id.to_s` or `user.to_param` to get string value of unique id of a document.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -115,54 +115,62 @@ def update_auth_header
...
+        puts "INVALID #{@resource.errors.full_messages}"
+        @resource = resource_class.find(@resource.to_param) # errors are keep after reload
+        # if we left the model in a bad state, something is wrong in our app
+        raise DeviseTokenAuth::Errors::InvalidModel, 'cannot set auth token in invalid model' unless @resource.valid?
```
dks17: Maybe put into exception's message information about errors `@resource.errors`? It will be more clearly for user. Something like this:
```ruby
raise DeviseTokenAuth::Errors::InvalidModel,
"cannot set auth token in invalid model. Errors: #{@resource.errors}" unless @resource.valid?
```

That do you think?

MaicolBen: the puts above are debugging purposes, the recent changes are breaking the tests

MaicolBen:
@dks17 It's ready, I had to move the test from the sessions controller to another one because it was failing only in travis for a random reason, and it was driving me crazy, the test makes sense in any controller that uses `update_auth_header`

dks17:
@MaicolBen ok. It looks good.
I would like that exception message test be added, but it's optional.

MaicolBen:
@dks17 thank you! it spotted a bug! I was outputting the ruby class name (array)

PR #1189, opened 2018-07-05
devise_mapping renders _users if namespace_name is an empty string

MitchellGeere:
replaced compact with reject(&:empty?) because compact only checks for nil but empty does both.


MaicolBen:
@MitchellGeere Sorry, but you are wrong, `nil` doesn't have an `empty?` method defined, check that in travis

PR #1195, opened 2018-07-25
feature: handling primary key type in generators

whois-marvin-42:
Migration generator handles if the primary key is a UUID (based on Devise implementation)


whois-marvin-42:
So much trouble with Travis for such a little modification

whois-marvin-42:
Thanks. I'm not gonna change it. I'll probably have more significant changes if I'm able to find the time to work on it, but I'll push a branch for each feature.

btw, talking about Travis (which was turning me crazy since all the builds were [passing](https://travis-ci.com/whois-marvin-42/devise_token_auth/branches) on my forked repository), I'm probably missing something because I think we can't rerun an individual job, it probably requires being part of the project's team (that's the _killing_ feature I was expecting and why I had to _fake_ changes ;)

MaicolBen:
sorry if you don't have the permission to rerun an individual job!

PR #1196, opened 2018-07-27
Add raise: false to skip_after_action :update_auth_header

arthuroliver:
My application has a controller that inherits from
DeviseTokenAuth:PasswordsController (for internal reasons).
When updating to Rails 5, I am getting the following error:

"After process_action callback :update_auth_header has not been defined"

Due to Rails 5 ActiveSupport::Callbacks#skip_callback now raising an
ArgumentError in case an unrecognized callback is undefined.

Rails 5 will raise the exception
https://github.com/rails/rails/blob/6dfab475ca230dfcad7a603483431c8e7a8f908e/activesupport/lib/active_support/callbacks.rb#L634

A solution is then to add the option raise: false, to the
skip_after_action :update_auth_header

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -3,7 +3,7 @@
 module DeviseTokenAuth
   class PasswordsController < DeviseTokenAuth::ApplicationController
     before_action :set_user_by_token, only: [:update]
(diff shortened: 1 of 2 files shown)


MaicolBen:
@arthuroliver why don't you have the callback defined? the test suite is running against rails 5

PR #1197, opened 2018-08-01
Add Factorybot

dks17:
Add factorybot factories instead of fixtures.


gemfiles/rails_5_1.gemfile
```diff
@@ -1,41 +1,40 @@
...
-  gem 'minitest'
-  gem 'minitest-focus'
-  gem 'minitest-rails'
-  gem 'minitest-reporters'
-  gem 'mocha', '>= 1.5'
-  gem 'pry'
-  gem 'pry-remote'
-
-  gem 'rubocop', require: false
+  gem "attr_encrypted"
```
MaicolBen: why going back with double quotes? we fixed this with rubocop

dks17: `bundle exec appraisal install` generates gem sets. I think the command is the reason. I could replace double quotes with editor. What do you think?

MaicolBen: well if it will happen all the time due to be autogenerated, leave as it is. I will continue reviewing this

PR #1206, opened 2018-08-24
Reset passwords stay valid until the password is reset

typhoon2099:
Or until the validity period has expired. This reverts an issue caused by #1008 where users who follow the password reset link from an email multiple times, or those who have anti spam filters which check links, are unable to reset their passwords.

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -44,7 +44,7 @@ module DeviseTokenAuth
     # this is where users arrive after visiting the password reset confirmation link
     def edit
       # if a user is not found, return nil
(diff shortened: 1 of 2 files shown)


MaicolBen:
I am not sure about this, but it makes sense, can you add a test checking the reset password token is invalid after the password is changed?

typhoon2099:
Is this okay to merge in yet?

mcelicalderon:
@typhoon2099 @MaicolBen I think this one is not actually right even with the test. The assertion is for a nil reset password token, but in the test, it was always nil, right? It should be tested with and without recoverable enabled and with a previously set `reset_password_token`. Would be happy to send a quick PR if what I say makes sense to you both.

mcelicalderon:
My bad, I tested with master and everything works as expected. Not sure where removing the password reset token happens :/

PR #1208, opened 2018-09-07
Add required_ruby_version

masatooba:
Because used hash literal that symbol key followed by a colon.

e.g.
https://github.com/masatooba/devise_token_auth/blob/add-required-ruby-version/lib/devise_token_auth/engine.rb#L41

The literal is not supported before 2.2.0.
Even if it is a version earlier than 2.2.0 you can install it, so when trying to use, raise error.

### DIFF
--- a/devise_token_auth.gemspec
+++ b/devise_token_auth.gemspec
@@ -20,6 +20,8 @@ Gem::Specification.new do |s|
   s.test_files = Dir['test/**/*']
   s.test_files.reject! { |file| file.match(/[.log|.sqlite3]$/) }
 
+  s.required_ruby_version = ">= 2.2.0"
+
   s.add_dependency 'rails', '< 6'
   s.add_dependency 'devise', '> 3.5.2', '< 4.6'
 


dks17:
This gem tested against `2.2.10` version. May be better set version more exactly?

https://github.com/lynndylanhurley/devise_token_auth/blob/2413d0015fa137ccc87444c0b2dbfd73c5f131f5/.travis.yml#L7-L11

@masatooba what do you think?

masatooba:
@dks17
Thank you.

This gem can be used if it is 2.1.0 or higher.
I think that I should correct the version specified by travis to 2.1.0 so that users currently using version 2.1.0 or higher and less than 2.1.10 can use it as it is.

dks17:
@masatooba 
I think you should consider increasing `ruby` version. Because of [Support of Ruby 2.2 has ended](https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/). And I think that support version `~> 2.2` for the gem will be deprecated in a while.

masatooba:
@dks17 
Yes. It's exactly as you say. I should consider increasing ruby version 😓 
In this case, set the ruby version to `2.1.10` and changing it when deprecated in the future?

dks17:
@masatooba Yes. I think so.

Also we should not forget about `devise` dependency. Devise still supports `2.1` ruby:
https://github.com/plataformatec/devise/blob/3b0bc08ec67dd073ddd6d043c71646c2784ced6c/.travis.yml#L3-L9

@MaicolBen Can we deprecated support for ruby `2.2` at the current time? It will decrease amount of Travis jobs in relation to #1209 pr. What do you think?

MaicolBen:
@dks17 that isn't a good reason to remove 2.2, I think if we come up with some blocker in code change that 2.2 doesn't allow us, we can remove it. This PR basically doesn't let people install it with 2.1 or lower, I'm ok with that

dks17:
@masatooba
Just question. Will this gem work with JRuby if you define `required_ruby_version` attribute?

masatooba:
@MaicolBen 
Thank you for review.

@dks17 
Yes. Each JRuby version matches a MRI version.
Jruby 9.0.0.0 matches MRI 2.2.2, so `required_ruby_version` will work.
The earlier Jruby version (e.g 1.7.27) does not support MRI 2.2.X, so it does not work.

PR #1049, opened 2017-12-22
Remove verbose warnings in test output

asartalo:
A fix for issue #976.

This removes most of the verbose warnings shown when running `rake test`.

PR #1217, opened 2018-09-22
Removed ensure pristine error due to not being used

### LINKED ISSUE #1135: ensure_pristine_resource error
v0.1.43

I'm having issues with the `ensure_pristine_resource` method in the lib's `app > controllers > devise_token_auth > concerns > set_user_by_token.rb`. Namely line 33:

```
@resource.assign_attributes(changes) if changes
```

I think I see what you're doing here, but `assign_attributes` is creating erroneous behavior when applying the hashed changes. For example:

```
irb(main)> u = User.first
=> #<User id: 1, name: "">
irb(main)> u.name = "John"
=> "John"
irb(main)> changes = u.changes
=> {"name"=>["", "John"]}
irb(main)> u.assign_attributes(changes)
=> nil
irb(main)> u
=> #<User id: 1, name: "[\"\", \"John\"]">
```

When using certain libraries, like Carrierwave, the application breaks at that line because I guess the parsing is off. I haven't done extensive testing, but doing something like this seems to at least parse the `changes` hash correctly:

```
changes.each { |attr, val| @resource[attr] = val[1] } if changes
```

Thoughts?

MaicolBen:
Resolves #1135

Trying to fix this issue, I realized that wasn't ever entering to the `if @resource.changed?` because now we are reloading the resource with #1188, but is that correct? we might lose changes! Well the thing is that we don't make changes to the resource between the beginning of `update_auth_header` and the call of `refresh_headers`, we can have changes between the developer's controller action and the call of `update_auth_header`  but that's bad, you shouldn't leave the model without saving.

`ensure_pristine_resource` was added in #1049 to suppress warnings, but now without it, we don't have them anymore due to the `reload` mentioned previously.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -28,7 +28,7 @@ bundle exec guard
 
 Guard will re-run each test suite when changes are made to its corresponding files.
 
-To run **just one test**: Flavio Castelli blogged about [how to execute a single unit test (or even a single test method)](https://flavio.castelli.me/2010/05/28/rails_execute_single_test/) instead of running the complete unit test suite.
+To run **just one test**: Flavio Castelli blogged about [how to execute a single unit test (or even a single test method)](https://flavio.castelli.me/2010/05/28/how-to-run-a-single-rails-unit-test/) instead of running the complete unit test suite.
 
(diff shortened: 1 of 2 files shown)


MaicolBen:
can you @dks17 review this one?

PR #1209, opened 2018-09-07
Add Mongoid support

dks17:
Add `Mongoid` support. See **Mongoid support** #1198 
In additional it fixes **codeclimate-test-reporter soon be deprecated** #1080

PR #1218, opened 2018-09-25
Test against Rails 5.2

krzysiek1507:


dks17:
@krzysiek1507 I have added test against `rails 5.2` in #1209 PR. Please check it.

PR #1237, opened 2018-11-06
Configure code climate

MaicolBen:


MaicolBen:
I will do it in the PR affected

PR #1242, opened 2018-12-07
Delete namespece and fix file name change to prevent override behavior of the default email validator

ihatov08:
I fixed DeviseTokenAuth::EmailValidator overrides the behavior of rails email validator.

### DIFF
--- a/app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
+++ b/app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
@@ -5,7 +5,7 @@ module DeviseTokenAuth::Concerns::UserOmniauthCallbacks
 
   included do
     validates :email, presence: true,if: :email_provider?
-    validates :email, 'devise_token_auth/email' => true, allow_nil: true, allow_blank: true, if: :email_provider?
+    validates :email, :devise_token_auth_email => true, allow_nil: true, allow_blank: true, if: :email_provider?
     validates_presence_of :uid, unless: :email_provider?
 
     # only validate unique emails among email registration users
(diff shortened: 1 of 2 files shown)


app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
```diff
@@ -5,7 +5,7 @@ module DeviseTokenAuth::Concerns::UserOmniauthCallbacks
 
   included do
     validates :email, presence: true,if: :email_provider?
-    validates :email, 'devise_token_auth/email' => true, allow_nil: true, allow_blank: true, if: :email_provider?
+    validates :email, :devise_token_auth_email => true, allow_nil: true, allow_blank: true, if: :email_provider?
```
MaicolBen: What does it fixes specifically? Because this was introduced for conflicts with other lib: #1233

MaicolBen: @skycocker will this conflict with solidus/spree ?

skycocker: yes, this will be a problem with pretty much any other lib which has a validator called `EmailValidator`. Maybe I introduced another bug I wasn't aware of? But it's hard to tell from just this revert.

ihatov08: Conflict EmailValidator in my rails app.

MaicolBen: I didn't receive a response so I'm going to merge this

Issue #1079, opened 2018-02-02
DB serialization issue in Rails 5.2.0


Marinlemaignan:
Unfortunately, upgrading to `Rails 5.2.0` (precisely Rails 5.2.0.rc1) with `devise_token_auth (0.1.43.beta1)` is impossible.

I cannot even manage to migrate my test database anymore.

* **error_message:**
`ActiveRecord::AttributeMethods::Serialization::ColumnNotSerializableError`
```
Column `tokens` of type ActiveRecord::Type::Json does not support `serialize` feature.
Usually it means that you are trying to use `serialize`
on a column that already implements serialization natively.
``` 

* **culprits:**
```
def tokens_has_json_column_type?
  database_exists? && table_exists? && self.columns_hash['tokens'] && self.columns_hash['tokens'].type.in?([:json, :jsonb])
end
```
```
unless tokens_has_json_column_type?
  serialize :tokens, JSON
end
```

* **stacktrace:**
```
Caused by:
ActiveRecord::AttributeMethods::Serialization::ColumnNotSerializableError: Column `tokens` of type ActiveRecord::Type::Json does not support `serialize` feature.
Usually it means that you are trying to use `serialize`
on a column that already implements serialization natively.
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_methods/serialization.rb:74:in `block in serialize'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:73:in `block in apply'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:72:in `each'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:72:in `inject'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:72:in `apply'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:53:in `block in load_schema!'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:52:in `each'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/attribute_decorators.rb:52:in `load_schema!'
(...)/gems/activerecord-5.2.0.rc1/lib/active_record/model_schema.rb:458:in `block in load_schema'
[...]

Issue #1240, opened 2018-11-26
no implicit conversion of Hash into String


fadyZohdy:
Hi we are currently hitting this issue with 
jruby 9.2.4.0 , rails 5.0.2 and devise_auth_token current master [commit](https://github.com/lynndylanhurley/devise_token_auth/commit/44b4a185479c9b1feb4e8854fda959f1b89c418f)
the same issue also used to happen before with release 1.0 but the stack trace was a bit different.
stack trace of current master
`
json/ext/Parser.java:170→ initialize
json/ext/Parser.java:142→ new
json-2.0.3-java/lib/json/common.rb:156→ parse
activesupport-5.0.2/lib/active_support/json/decoding.rb:21→ decode
activerecord-5.0.2/lib/active_record/coders/json.rb:9→ load
activerecord-5.0.2/lib/active_record/type/serialized.rb:18→ deserialize
/opt/jruby/lib/ruby/stdlib/delegate.rb:349→ block in delegating_block
activerecord-5.0.2/lib/active_record/type/serialized.rb:18→ deserialize
/opt/jruby/lib/ruby/stdlib/delegate.rb:349→ block in delegating_block
activerecord-5.0.2/lib/active_record/type/serialized.rb:18→ deserialize
activerecord-5.0.2/lib/active_record/attribute.rb:140→ type_cast
activerecord-5.0.2/lib/active_record/attribute.rb:38→ value
activerecord-5.0.2/lib/active_record/attribute_set.rb:40→ fetch_value
activerecord-5.0.2/lib/active_record/attribute_methods/read.rb:62→ _read_attribute
activerecord-5.0.2/lib/active_record/attribute_methods/read.rb:36→ __temp__47f6b656e637
[PROJECT_ROOT]/vendor/cache/devise_token_auth-44b4a185479c/app/models/devise_token_auth/concerns/active_record_support.rb:32→ set_empty_token_hash
activesupport-5.0.2/lib/active_support/callbacks.rb:382→ block in make_lambda
activesupport-5.0.2/lib/active_support/callbacks.rb:207→ block in halting_and_conditional
activesupport-5.0.2/lib/active_support/callbacks.rb:456→ block in call
org/jruby/RubyArray.java:1789→ each
activesupport-5.0.2/lib/active_support/callbacks.rb:456→ call
activesupport-5.0.2/lib/active_support/callbacks.rb:101→ __run_callbacks__
[...]

Issue #1224, opened 2018-10-05
undefined method '[]' for nil:NilClass when confirming email


moray95:
Hello,

I am getting a `undefined method '[]' for nil:NilClass` error when confirming an email. The issue appears very rarely and requires some luck to reproduce. I have noticed the issue appears because the user's tokens doesn't contain the given client id in the line `tokens[args[:client_id]]['expiry']`. I have been able to reproduce the issue locally (again with very low reproducibility rate) by clicking multiple times to the confirmation link sent by mail. You can find additional information below.

* **Version**:  0.2.0
* **Request and response headers**: http://localhost:3000/auth/confirmation?config=default&confirmation_token=XXX&redirect_url=https%3A%2F%2Fexample.com
* **Rails Stacktrace**: 
```
vendor/bundle/ruby/2.5.0/gems/devise_token_auth-0.2.0/app/models/devise_token_auth/concerns/user.rb:211:in `build_auth_url'
vendor/bundle/ruby/2.5.0/gems/devise_token_auth-0.2.0/app/controllers/devise_token_auth/confirmations_controller.rb:26:in `show'
vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/abstract_controller/base.rb:194:in `process_action'
vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/action_controller/metal/rendering.rb:30:in `process_action'
vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.1/lib/abstract_controller/callbacks.rb:42:in `block in process_action'
vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/callbacks.rb:109:in `block in run_callbacks'
vendor/bundle/ruby/2.5.0/gems/sentry-raven-2.7.4/lib/raven/integrations/rails/controller_transaction.rb:7:in `block in included'
vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/callbacks.rb:118:in `instance_exec'
vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.1/lib/active_support/callbacks.rb:118:in `block in run_callbacks'
[...]

PR #1250, opened 2019-01-20
Tokens serialization

dks17:
This is custom [coder](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) for tokens column with known object's structure (only Hash). This is based on `JSON` coder with some edge cases (nil, blank string).

`PG` registers `jsonb` as alias type for `json` and uses own [text coder](https://github.com/ged/ruby-pg/blob/b920bb33eaa4797cdfdca8885dd56dbd04d6b8ea/lib/pg/basic_type_mapping.rb#L218-L219) where also uses `JSON` coder and few other date/time formats for [encoding](https://github.com/ged/ruby-pg/blob/1d9883d06bdf5413880f20bbb445720dedebd736/lib/pg/text_encoder.rb)/[decoding](https://github.com/ged/ruby-pg/blob/1d9883d06bdf5413880f20bbb445720dedebd736/lib/pg/text_decoder.rb).

Should help to serialize/deserialize tokens properly.

This should fix:
#1079 #121
and may:
#1240 #1224 

Need help testing the PR before it will be merged.

**UPDATE**

- Updated `dump` method.
- Deleted code from `load` method that recognize case when json is blank string (it is unlikely case).

### DIFF
--- a/app/models/devise_token_auth/concerns/tokens_serialization.rb
+++ b/app/models/devise_token_auth/concerns/tokens_serialization.rb
@@ -1,0 +1,19 @@
+module DeviseTokenAuth::TokensSerialization
+  # Serialization hash to json
+  def self.dump(object)
+    object.each_value(&:compact!) unless object.nil?
+    JSON.generate(object)
+  end
+
+  # Deserialization json to hash
+  def self.load(json)
+    case json
+    when String
+      JSON.parse(json)
+    when NilClass
+      {}
+    else
+      json
+    end
+  end
+end
--- a/app/models/devise_token_auth/concerns/active_record_support.rb
+++ b/app/models/devise_token_auth/concerns/active_record_support.rb
@@ -1,12 +1,10 @@
+require_relative 'tokens_serialization'
+
 module DeviseTokenAuth::Concerns::ActiveRecordSupport
   extend ActiveSupport::Concern
 
   included do
-    serialize :tokens, JSON unless tokens_has_json_column_type?
-
-    # can't set default on text fields in mysql, simulate here instead.
-    after_save :set_empty_token_hash
-    after_initialize :set_empty_token_hash
+    serialize :tokens, DeviseTokenAuth::TokensSerialization
   end
 
   class_methods do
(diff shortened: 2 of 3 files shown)


app/models/devise_token_auth/concerns/tokens_serialization.rb
```diff
@@ -0,0 +1,18 @@
+module DeviseTokenAuth::TokensSerialization
+  # Serialization hash to json
+  def self.dump(object)
+    object.nil? ? '{}' : object.each_value(&:compact!).to_json
```
dks17: I noticed that value of `last_token` very often equals nil (it seems batch) while `updated_at` is still present and updated. So delete (`&:compact!`) all keys with nil value.

dks17: We could store 'null' instead of '{}' to `NULL` operators could be used.

dks17:
@MaicolBen Wrote small description about it at the top message.

app/models/devise_token_auth/concerns/tokens_serialization.rb
```diff
@@ -0,0 +1,19 @@
+module DeviseTokenAuth::TokensSerialization
+  # Serialization hash to json
+  def self.dump(object)
+    object.each_value(&:compact!) unless object.nil?
```
dks17: If we care about storing `null` values we can delete this line.
See `last_token` below:

"{\"nlw47IfMSJxHl8v2bvhrOw\":{\"token\":\"$2a$10$3yWVUhPtnn/vigTztah2fuioMMEB6R8aZunkFOjzguV1ZCZ8kga16\",\"expiry\":1549026407,\"last_token\":null,\"updated_at\":\"2019-01-18T13:06:32.876Z\"}}"

app/models/devise_token_auth/concerns/tokens_serialization.rb
```diff
@@ -0,0 +1,19 @@
...
+    JSON.generate(object)
```
dks17: Serialize nil to `null`:
```ruby
JSON.generate(nil)
=> "null"
```

app/models/devise_token_auth/concerns/tokens_serialization.rb
dks17: Anyway `nil` or `{}` is stored as `null`.

kdubel:
Thanks for this fix! I had v1.1.0 of gem and the problem existed, updated to your fix and everything works fine. @MaicolBen maybe it's a good idea to bump version as searching for this issue takes some time?

softwaregravy:
I'm tracing through the threads on this issue as well. I agree it would be great to get this out.

PR #1252, opened 2019-01-24
DeviseTokenAuth::Url.generate should accept relate path

mengqing:
When a relative url or empty string is passed to `DeviseTokenAuth::Url.generate`, it should parse and return the same relative url back.

### DIFF
--- a/lib/devise_token_auth/url.rb
+++ b/lib/devise_token_auth/url.rb
@@ -4,15 +4,10 @@ module DeviseTokenAuth::Url
 
   def self.generate(url, params = {})
     uri = URI(url)
+    query_params = Hash[URI.decode_www_form(uri.query || '')].merge(params)
+    uri.query = URI.encode_www_form(query_params)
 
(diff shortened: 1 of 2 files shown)


mengqing:
Hi @MaicolBen, I'm not exactly sure why the spec failed. I've tried testing them locally and they all passed. Do you have any idea on why the spec is failing?

lib/devise_token_auth/url.rb
```diff
@@ -4,15 +4,10 @@ module DeviseTokenAuth::Url
 
   def self.generate(url, params = {})
     uri = URI(url)
+    query_params = Hash[URI.decode_www_form(uri.query || '')].merge(params)
```
MaicolBen: The problem is that `expiry` gets lost here in rails 4.2 (`/controllers/devise_token_auth/passwords_controller_test.rb:207`). I thought `URI` is from ruby, so it's weird

PR #1253, opened 2019-01-24
User#send_reset_password_instructions redirect_url not set as default if blank

mengqing:
`redirect_url` should be set to `DeviseTokenAuth.default_password_reset_url` if not present in the options

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -69,6 +69,7 @@ module DeviseTokenAuth::Concerns::User
 
       # fall back to "default" config name
       opts[:client_config] ||= 'default'
+      opts[:redirect_url] ||= DeviseTokenAuth.default_password_reset_url
 
       send_devise_notification(:reset_password_instructions, token, opts)
       token


test/controllers/devise_token_auth/passwords_controller_test.rb
```diff
@@ -330,8 +330,7 @@ class DeviseTokenAuth::PasswordsControllerTest < ActionController::TestCase
           DeviseTokenAuth.default_password_reset_url = @redirect_url
```
MaicolBen: `@redirect_url` isn't null, why this change in the spec ?

mengqing: The test was to test "Using default_password_reset_url", however previously the redirect url was passed as a param, and that's why the spec passed, and thus the change in the spec to show that the test is still valid without the url being passed to the controller

MaicolBen: and how the spec in line 61 passes? we have a validation for the missing redirect url in the controller

MaicolBen: @mengqing ☝️

mengqing: @MaicolBen My bad, it looks like on https://github.com/lynndylanhurley/devise_token_auth/blob/f39244eabcfd58f0fa31421774fd10de569baea4/app/controllers/devise_token_auth/passwords_controller.rb#L181, `DeviseTokenAuth.default_password_reset_url` is already used as a backup for `@redirect_url`, hence the spec in line 61 passes.

The issue I wanted to fix in this PR is really for `resource.send_reset_password_instructions` where the method would still work if no redirect_url is passed as an option, eg: calling it directly. Since there isn't a test to cover `send_reset_password_instructions`, do I need to create a new test for this?

MaicolBen: If you can, add a spec for this case in the model

PR #1258, opened 2019-02-12
devise_token_group does return the header part (CID,token, client-id ETC)

williamhatch:
I found it is the issue in @resource.is_a?(rc) when devise_token_group


williamhatch:
OK add more detail here:
in appicaiton_controller.rb
```
devise_token_auth_group :member, contains: [:user, :company]
before_action :authenticate_member! # Ensure someone is logged in
skip_before_action :authenticate_member!, if: :devise_controller?
```
Then use postman client to test it, user return token etc but company doesnt!

PR #1261, opened 2019-02-26
Upgrade gem dependency (Devise)

ngpestelos:
Upgrades Devise to allow Rails applications running on Devise v4.6 to use this gem.


MaicolBen:
I will close it in favor of #1269 as it made the specs pass

PR #1266, opened 2019-03-12
Convert q_value to_s

pioz:
Ensure that `q_value` is a string to call method `downcase!` (line 12) and `strip!`. You can pass for example the email parameter = null with Postman: in this scenario you get 500 `NoMethodError (undefined method `downcase!' for nil:NilClass)`


pioz:
Any update?

PR #1269, opened 2019-03-14
allow devise < 4.7

doits:
Tests pass and my app works with it, too, but maybe someone with a more complicated setup could test it a little bit more.

Note: I had to explicitly `require 'devise/orm/active_record'` in `config/initializers/devise_token_auth.rb` or else I had some errors about missing method `devise` in my models. Before I didn't had it and everything worked, but maybe now it is required? Not sure if this should be added to `lib/generators/devise_token_auth/templates/devise_token_auth.rb` here, too.

(Devise < 4.6 has a [security issue](https://github.com/plataformatec/devise/issues/4981), so I think it should be updated as fast as possible)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ bundler_args: --without development
 rvm:
   - 2.3.8
   - 2.4.5
-  - 2.5.3
-  - 2.6.0
+  - 2.5.5
+  - 2.6.2
 
 gemfile:
   - gemfiles/rails_4_2.gemfile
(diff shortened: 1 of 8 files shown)


doits:
I hope you don't mind I updated the ruby versions in CI, fixed the sqlite3 issue and added a CI test against rails 6.0.beta (except mongoid since I think it does not support rails 6 yet) in this MR, too. 🤞 that tests pass now.

Gemfile
```diff
@@ -22,6 +22,7 @@ group :development, :test do
   gem 'omniauth-github',        git: 'https://github.com/intridea/omniauth-github'
   gem 'omniauth-google-oauth2', git: 'https://github.com/zquestz/omniauth-google-oauth2'
   gem 'rack-cors', require: 'rack/cors'
+  gem 'sqlite3', '~> 1.3.6'
```
MaicolBen: why adding sqlite?

.travis.yml
```diff
@@ -70,10 +70,14 @@ matrix:
           ./cc-test-reporter upload-coverage;
         fi
   exclude:
-    - rvm: 2.6.0
+    - rvm: 2.6.2
       gemfile: gemfiles/rails_4_2.gemfile
   fast_finish: true
 
+before_install:
+  - "[[ $BUNDLE_GEMFILE == *rails_4_2* ]] && gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true"
```
MaicolBen: why this change?

Gemfile
doits: had trouble running specs on my dev machine, but can be removed of course if you don't see it fit.

```suggestion
```

.travis.yml
doits: because [bundler 2 does not play well with rails 4.2](https://travis-ci.org/lynndylanhurley/devise_token_auth/jobs/506927792)

Gemfile
MaicolBen: if you can, try to remove it from here and other places, if you have issues that can happen to other developers, we can keep it.

doits: I removed it now, tests work fine without it now, too. I think it is because the sqlite version is specified in other gemfiles now. In them it is needed though, since otherwise it will install sqlite >= 1.4 which breaks with current rails versions.

doits:
I think now it's ready with only the minimal changes needed.

dkniffin:
Can we get a patch version bump for this?

MaicolBen:
@dkniffin @Polidoro I released v1.1.0 which includes some features and bug fixes (e.g. Mongo support), if you need it in v1.0.0 let me know and I can create a PR for that. If you were using master, you won't have any issue.

You can see the differences in https://github.com/lynndylanhurley/devise_token_auth/compare/v1.0.0...master

PR #1284, opened 2019-04-19
Enhacenment/rubocop fixes

GAKINDUSTRIES:
This PR try to address most of Rubocop issues. It gets the offenses from 676 down to 81, but I think that there were too many changes in this PR, so I will try to address the others in a separate PR.

Each correction for a given cop was made in a separate commit, so to make reviewing easier. I can squash the commits when approved if you want.
Please take your time to review it since I did a lot of changes in it.

After all, warnings are fixed, I would like to propose including the linter in CI and update the contribution section so every new PR does not add any warnings.

In regards of Style/AlignHash cop, it would be good to discuss if we prefer to use `# EnforcedHashRocketStyle: table` or `# EnforcedHashRocketStyle: key (default)`. I use the default one since it was not set in the configuration file, so I assume that you guys prefer to use that one, but we can change it if you want.


MaicolBen:
@GAKINDUSTRIES can you rebase? sorry!

PR #1296, opened 2019-06-02
Relax bcrypt version

MaicolBen:
This happened to me when I was upgrading devise and rails

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,11 +17,11 @@ gemspec
 
 group :development, :test do
   gem 'attr_encrypted'
-  gem 'figaro', git: 'https://github.com/laserlemon/figaro'
-  gem 'omniauth-facebook', git: 'https://github.com/mkdynamic/omniauth-facebook'
-  gem 'omniauth-github',        git: 'https://github.com/intridea/omniauth-github'
-  gem 'omniauth-google-oauth2', git: 'https://github.com/zquestz/omniauth-google-oauth2'
-  gem 'rack-cors', require: 'rack/cors'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -17,11 +17,11 @@ gemspec
 
 group :development, :test do
   gem 'attr_encrypted'
-  gem 'figaro', git: 'https://github.com/laserlemon/figaro'
-  gem 'omniauth-facebook', git: 'https://github.com/mkdynamic/omniauth-facebook'
-  gem 'omniauth-github',        git: 'https://github.com/intridea/omniauth-github'
-  gem 'omniauth-google-oauth2', git: 'https://github.com/zquestz/omniauth-google-oauth2'
-  gem 'rack-cors', require: 'rack/cors'
+  gem 'figaro'
```
MaicolBen: We don't need Github here (the specs are green locally) and makes `bundle` slow

devise_token_auth.gemspec
```diff
@@ -22,9 +22,9 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.2.0"
 
-  s.add_dependency 'rails', '>= 4.2.0', '< 6'
+  s.add_dependency 'rails', '>= 4.2.0', '< 6.1'
   s.add_dependency 'devise', '> 3.5.2', '< 4.7'
-  s.add_dependency 'bcrypt', '3.1.12'
```
MaicolBen: now it has the same version restriction than devise

Issue #1301, opened 2019-06-10
Huge performance downgrade from v0.1.43 to v1.1.0


LeoValverde:
When I upgraded from v0.1.43 to v.1.1.0, the overall performance of the controller got worse by a factor of 3.

In the picture below, you can see the response times as bars. Response time is the Y axis, and the period of the day is the X axis.

The controller times are the dark purple portion of the bar. It was valued around 120ms before the upgrade, but it increased to 400ms right after the deploy.

Is there some explanation to this raise? I'm going to rollback this deploy, but I hope we can figure this out.

![image](https://user-images.githubusercontent.com/10124568/59226442-63b43880-8ba9-11e9-93c2-eced69955bd4.png)

Thank you.

lynndylanhurley:
@LeoValverde thanks for the report. I'll look into this ASAP

LeoValverde:
Great! When I was rolling back, this was the gemfile.lock diff.

After the rollback, the performance was solved.

![image](https://user-images.githubusercontent.com/10124568/59276634-154c7b80-8c35-11e9-9f74-960ad1671faf.png)

lynndylanhurley:
That's really helpful. Thanks @LeoValverde

LeoValverde:
My bad @lynndylanhurley, this was not the rollback that solved the problem. Actually, everything slows down when we start using bcrypt `v3.1.13` instead of `v3.1.12`. I just forced my project to use `v3.1.12` and could successfully update devise_token_auth.

This update happened when I run` bundle install` to set up the `v1.1.0`, but maybe it's not something you can solve by yourself...

I'll confirm that was the problem, but I'm pretty confident. Just updating you to avoid you wasting your time. I'll be right back with the final answer.

Thx anyway :)

LeoValverde:
That was the problem! `devise_token_auth` depends on `devise`, which depends on `bcrypt`. The last one was upgraded to a slow version. I forced it not to upgrade and everything was solved. Thanks

booleanbetrayal:
This is likely the changeset -- 

https://github.com/codahale/bcrypt-ruby/pull/181

booleanbetrayal:
I've validated that changing the default cost back to `10` on BCrypt `3.1.13` seems to produce similar response times through DTA endpoints. Seems like the argument that the BCrypt `cost` value needs to be increased generally is sound, and we shouldn't override it.

[...]

PR #1305, opened 2019-06-14
chore(deps): update bcrypt-3.1.13

booleanbetrayal:
This update includes a bump to the default `cost` from `10` to `12`. Seems that we should support this standard going forward since the `cost` had not changed since initial release in 2007 (?).

see also: https://github.com/lynndylanhurley/devise_token_auth/issues/1301#issuecomment-501396488


booleanbetrayal:
Closed in favor of https://github.com/lynndylanhurley/devise_token_auth/pull/1296

Issue #1304, opened 2019-06-13
Token validation could raise exception in v1.1.0


LeoValverde:
The validation made on line 105 of the `set_user_by_token.rb` is causing me trouble.

There are some cases where my before_validation callbacks change my user model. I do it for cleaning some old dates from my model, before running validations that ensure these dates are clean.

After I upgraded to DTA v1.1.0, in some requests, the call `unless @resource.reload.valid?` leaves the `@resource` with unpersisted changes, and the exception is thrown right after, when the `@resource.with_lock` is called.

```
A RuntimeError occurred in token_validations#validate_token:

Locking a record with unpersisted changes is not supported. Use `save` to persist the changes, or `reload` to discard them explicitly.

```

I could change my callback to be an after save, but I really don't understand the goal of this validation (and the reload, that causes an extra database query), because we lock the object right after it, which is when we query for the updated version of the object.

Tiltorito:
Hi, 

are you sure that's the case? because on line 106 it reloads the model completely and it should be on a clean state and ready to obtain the lock.

LeoValverde:
Agreed!

Tiltorito:
@LeoValverde do you know if the project is still maintained? Because my PR and our issues haven't been addressed at all yet :P

zachfeldman:
@Tiltorito @LeoValverde there was an effort to revitalize this project about a year ago by myself, @lynndylanhurley , and @MaicolBen . I think we've all gotten pretty busy again and have lost the time, but there's still $500 in our Open Collective account:
https://opencollective.com/devise_token_auth

I'm sure Lynn wouldn't mind someone else helping out/taking some reigns, but I can't speak for him and he's the author.

MaicolBen:
Yep, we need collaborators! I only have time to review PRs 😞

Tiltorito:
@zachfeldman @MaicolBen I would like to contribute to the project. I am not sure if I will be able to contribute that much on the code because I am a bit inexperienced but I would happily help with the documentation

Issue #1302, opened 2019-06-11
Too many database calls


Tiltorito:
Way too many database calls. This is happening on every endpoint, but I'm using /validate_token as an example. Is this the expected behaviour to make 3 database calls just to fetch the user?

Those many database calls lead to other issues, for example, 500 error responses because user table is locked

![Screen Shot 2019-06-11 at 20 03 04](https://user-images.githubusercontent.com/31311538/59299189-f45f4700-8c83-11e9-9dd9-84b35bbde58f.png)

Tiltorito:
Hi, 

I am not sure if it is going to help but, I set config.change_headers_on_each_request = false and reduced the database UserLoad calls for a user down to 2 from 3.

I debugged my application and seems the first calls is coming from my before_action :authenticate_user! which makes sense.

The second call, however, is coming from [set_user_by_token.rb](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb) line 97 where reload is called.

Why is that? Do we need this kind of check?

LeoValverde:
Great, @Tiltorito ! It is painful to see a redundant database call on EVERY request.

LeoValverde:
This validation made on line 105 is causing me trouble.

There are some cases where my `before_validation` callbacks change my user model. I do it for cleaning some old dates from my model, before running validations that ensure these dates are clean.

After I upgraded to DTA v1.1.0, in some requests, the call `unless @resource.reload.valid?` left the `@resource` with unpersisted changes, and the exception was thrown right after, when the `@resource.with_lock` is called.

I could change my callback to be an after save, but I really don't understand the goal of this validation, because we lock the object right after it, which is when we query for the updated version of the object.

PR #1308, opened 2019-06-19
removed redundant db calls

Tiltorito:
Removed redundant DB calls. This PR address issues #1302 #1304

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -94,7 +94,7 @@ module DeviseTokenAuth::Concerns::SetUserByToken
     if @used_auth_by_token && !DeviseTokenAuth.change_headers_on_each_request
       # should not append auth header if @resource related token was
       # cleared by sign out in the meantime
-      return if @resource.reload.tokens[@token.client].nil?
+      return if @resource.tokens[@token.client].nil?
 
       auth_header = @resource.build_auth_header(@token.token, @token.client)
 
@@ -102,7 +102,7 @@ module DeviseTokenAuth::Concerns::SetUserByToken
       response.headers.merge!(auth_header)
 
     else
-      unless @resource.reload.valid?
+      unless @resource.valid?
         @resource = resource_class.find(@resource.to_param) # errors remain after reload
         # if we left the model in a bad state, something is wrong in our app
         unless @resource.valid?
@@ -116,6 +116,11 @@ module DeviseTokenAuth::Concerns::SetUserByToken
   private
 
   def refresh_headers
+    # At this point restore the attributes so we can obtain the lock
+    # otherwise an exception is thrown while trying to obtain the lock
+    # with unsaved data on the resource.
+    @resource.restore_attributes
+
     # Lock the user record during any auth_header updates to ensure
     # we don't have write contention from multiple threads
     @resource.with_lock do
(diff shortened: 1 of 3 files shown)


MaicolBen:
They were in purpose but it seems you found a better workaround, can you fix the failed specs?

Tiltorito:
> They were in purpose but it seems you found a better workaround, can you fix the failed specs?

Hi, thanks for your comment. 

Yes, I can fix it, the problem is I can't run the tests locally :(.

Could you please give me instructions on how?

I use: 

1. rake app:db:load_config
2. rails app:environment
3. bundle exec rake

and I am getting  Could not find table 'mangs' or 'users'.

If on the other hand, I run  rake app:test:db I get Connection refused - connect(2) for "localhost" port 1025

MaicolBen:
If you have issues setting up a database, you can just use sqlite, you can follow https://github.com/lynndylanhurley/devise_token_auth/blob/master/.github/CONTRIBUTING.md#run-the-tests

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -94,15 +94,15 @@ def update_auth_header
     if @used_auth_by_token && !DeviseTokenAuth.change_headers_on_each_request
       # should not append auth header if @resource related token was
       # cleared by sign out in the meantime
-      return if @resource.reload.tokens[@token.client].nil?
```
MaicolBen: This line fails a spec that was added here https://github.com/lynndylanhurley/devise_token_auth/pull/531/files#diff-41be446a925979b6d8df8a6361c2ef48R343

MaicolBen: Maybe you can split this PR in 2, the second line can fix #1304 but this one needs a different approach, I'm thinking that we might need a good refactor with a different idea to avoid going to the database twice.

Tiltorito: Hi, 

Thank you for your responses. I fixed the spec. The problem is, I think they had made the second query intentionally. So, if your request has been processed but you log out in the meantime using the same access token then don't return the access_token back to any requests that contained the token. Makes sense... but, I don't think its that big of a deal. Because, the token is removed from the DB at this point and even if that DID happen, the clients shouldn't be able to access any of the system's resources using the access_token that was returned to them. Hope I make sense.

MaicolBen: Makes sense, I would recommend splitting this in 2 PRs, so #1304 is addressed and will be in the next version.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -116,6 +116,11 @@ def update_auth_header
   private
 
   def refresh_headers
+    # At this point restore the attributes so we can obtain the lock
+    # otherwise an exception is thrown while trying to obtain the lock
+    # with unsaved data on the resource.
+    @resource.restore_attributes
```
MaicolBen: This doesn't exist in Mongo 😞

Tiltorito: :(

MaicolBen: Maybe leave the previous behavior for Mongo and this for rails. But split in 2 PRs so is easily done and we avoid breaking changes.

Tiltorito: Hi, I apologize for my late reply. How can I check if the database is going to be Mongo?

MaicolBen: `if const_defined?('Mongoid')`

PR #1315, opened 2019-07-17
Creating the 'previous_token' key to be in between the 'token' and th…

LeoValverde:
…e 'last_token' keys. Now there are 2 valid tokens for each user at any moment to prevent from data loss due to interrupted requests

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -116,6 +116,7 @@ module DeviseTokenAuth::Concerns::User
     # ghetto HashWithIndifferentAccess
     expiry     = tokens[client]['expiry'] || tokens[client][:expiry]
     token_hash = tokens[client]['token'] || tokens[client][:token]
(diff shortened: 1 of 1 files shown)


app/models/devise_token_auth/concerns/user.rb
```diff
@@ -125,7 +126,10 @@ def token_is_current?(token, client)
       DateTime.strptime(expiry.to_s, '%s') > Time.zone.now &&
 
       # ensure that the token is valid
-      DeviseTokenAuth::Concerns::User.tokens_match?(token_hash, token)
+      (
+        (token_hash.present? && DeviseTokenAuth::Concerns::User.tokens_match?(token_hash, token)) or
```
MaicolBen: I would move `DeviseTokenAuth::Concerns::User.tokens_match?` to a method here, so you don't repeat twice

MaicolBen:
Got overriden by https://github.com/lynndylanhurley/devise_token_auth/pull/1520

PR #1330, opened 2019-08-19
Skip callback when active record

enomotodev:
For ActiveRecord, `send_on_create_confirmation_instructions` method could not be skipped, so it can be skipped.

This is the target code on the devise side.

- https://github.com/plataformatec/devise/blob/master/lib/devise/models/confirmable.rb#L51-L57

### DIFF
--- a/app/controllers/devise_token_auth/registrations_controller.rb
+++ b/app/controllers/devise_token_auth/registrations_controller.rb
@@ -31,8 +31,9 @@ module DeviseTokenAuth
       return render_create_error_redirect_url_not_allowed if blacklisted_redirect_url?
 
       # override email confirmation, must be sent manually from ctrl
-      resource_class.set_callback('create', :after, :send_on_create_confirmation_instructions)
-      resource_class.skip_callback('create', :after, :send_on_create_confirmation_instructions)
+      callback_name = defined?(ActiveRecord) && resource_class < ActiveRecord::Base ? :commit : :create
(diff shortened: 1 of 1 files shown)


app/controllers/devise_token_auth/registrations_controller.rb
```diff
@@ -31,8 +31,13 @@ def create
       return render_create_error_redirect_url_not_allowed if blacklisted_redirect_url?
 
       # override email confirmation, must be sent manually from ctrl
-      resource_class.set_callback('create', :after, :send_on_create_confirmation_instructions)
-      resource_class.skip_callback('create', :after, :send_on_create_confirmation_instructions)
+      if defined?(ActiveRecord) && resource_class < ActiveRecord::Base
+        resource_class.set_callback('commit', :after, :send_on_create_confirmation_instructions)
```
MaicolBen: Ugh, this is ugly, why not the next?
```ruby
callback_name = defined?(ActiveRecord) && resource_class < ActiveRecord::Base ? :commit : :create
resource_class.set_callback(callback_name, :after, :send_on_create_confirmation_instructions)
resource_class.skip_callback(callback_name, :after, :send_on_create_confirmation_instructions)
```

enomotodev: @MaicolBen Thank you for the review.

That's certainly simpler! Look at it again because I fixed it.

MaicolBen: Cool, checking why travis is failing...

PR #1340, opened 2019-09-09
Fix locking with unsaved changes bug

mrkrlli:
@MaicolBen @Tiltorito

- This problem has been affecting me with `attr_encrypted` and `lockbox` gems, since they re-encrypt data on validation. 
- Breaking out the PR from here: https://github.com/lynndylanhurley/devise_token_auth/pull/1308/files
- Fixing the bug where calling `valid?` leaves unsaved changes, which throws an error when calling `@resource.with_lock`

### DIFF
--- a/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
+++ b/app/controllers/devise_token_auth/concerns/set_user_by_token.rb
@@ -116,6 +116,13 @@ module DeviseTokenAuth::Concerns::SetUserByToken
   private
 
   def refresh_headers
+    # At this point restore the attributes so we can obtain the lock
+    # otherwise an exception is thrown while trying to obtain the lock
+    # with unsaved data on the resource.
+    unless const_defined?('Mongoid') do
+      @resource.restore_attributes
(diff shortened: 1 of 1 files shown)


Tiltorito:
@mrkrlli Thank you very much for actually doing it, I couldn't find time to fix my PR :)

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -116,6 +116,11 @@ def update_auth_header
   private
 
   def refresh_headers
+    # At this point restore the attributes so we can obtain the lock
+    # otherwise an exception is thrown while trying to obtain the lock
+    # with unsaved data on the resource.
+    @resource.restore_attributes unless const_defined?('Mongoid')
```
MaicolBen: I think you have to use `defined?` instead, `const_defined?` is with `Module.const_defined?("something")`

mrkrlli:
@MaicolBen Made the change, sorry it took so long.

PR #1343, opened 2019-09-24
Fix to be able to use Devise::confirmable module

### LINKED ISSUE #1013: Using DeviseTokenAuth::Concerns::User breaks Devise::confirmable and Devise::reconfirmable
Culprits being :
```
    # don't use default devise email validation
    def email_required?
      false
    end

    def email_changed?
      false
    end

    def will_save_change_to_email?
      false
    end
```
why is that

makicamel:
## Summary

Fix #1013 .

### 1. Override `postpone_email_change?` method of devise

In DeviseTokenAuth, `will_save_change_to_email?` and `email_changed?` methods are
overriden to return `false` for disabling Devise default email validation.
This override breaks Devise confirmable module.
To fix this issue, override `postpone_email_change?`
not to use `email_changed?` and `will_save_change_to_email?`,
but to use `email_was` and `email_in_database`.
(I didn't change `will_save_change_to_email?` `email_changed?` methods to return `super`
because it breaks DeviseTokenAuth)

### 2. Create `send_confirmation_email flag` to use Devise confirmable

For backward compatible, not override `postpone_email_change?` in default.
Make it be able to use confirmable module by setting `config/initializers/devise_token_auth.rb`.


app/models/devise_token_auth/concerns/user.rb
```diff
@@ -44,6 +44,28 @@ def email_required?; false; end
     def email_changed?; false; end
     def will_save_change_to_email?; false; end
 
+    if Devise.rails51?
+      def postpone_email_change?
```
makicamel: > Similar blocks of code found in 2 locations. Consider refactoring.
Humm, I think keeping code like this helps to make it clear this method overrides Devise method that.
Which is better, keeping or refactoring?

ref. https://github.com/plataformatec/devise/blob/f48b6f1651233a39c52dd70e9ac6c35f70086062/lib/devise/models/confirmable.rb#L279-L299

MaicolBen: I would prefer moving this to a concern/module, this file has too much code and we should try to clean this lib as much as possible

makicamel: Thank you for reply, @MaicolBen !
I fully agree to try to clean this library!
I moved `postpone_email_change?` method to module, so please take a look again :smile_cat:

MaicolBen: thanks!

alexmreis:
@MaicolBen can you make a release with this fix in it please?

experimatt:
+1 I'm eager to use this once it's released

PR #1349, opened 2019-10-25
Fix dead link

NRauh:
I noticed that the usage tldr link just takes you to the main readme which doesn't have any usage info (at this time).

I was thinking it could also go here https://github.com/lynndylanhurley/devise_token_auth/tree/master/docs/usage but thought the gitbook docs looked nicer and the content is the same from what I could tell.


booleanbetrayal:
thanks @NRauh !

PR #1363, opened 2019-12-02
fix client nil

drougojrom:
In our case it happened that the client was an empty hash, because the user was authenticated from a different place. Although, in order to get the access to swagger, with was mounted into the api, we needed to go through the main Api controller.

That resolved in ```undefined method `[]=' for nil:NilClass```

The solution is quite simple and it's just a nil check


MaicolBen:
I think you should fix your client, why make the change here?

PR #1364, opened 2019-12-03
Don't rely on Time#to_s to return a precise updated_at

briandunn:
The project I'm on has an overridden `::Time::DATE_FORMATS[:default]` that does not include the seconds in the serialized time string. This broke batch detection in a way that was very difficult to track down. This PR calls `as_json` before serializing, just like the [default json serializer](https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activesupport/lib/active_support/json/encoding.rb#L35)


MaicolBen:
@briandunn can you fix travis?

PR #1366, opened 2019-12-05
Add rails 6.0 config to travis

brateq:
Added Travis config for Rails 6 and Ruby 2.7 for all databases.


brateq:
This is a good time to remove unsupported Ruby (2.3.8) and Rails (4.2, 5.0, 5.1) versions from Travis config. What do you think?

test/lib/generators/devise_token_auth/install_generator_test.rb
```diff
@@ -47,7 +47,7 @@ class InstallGeneratorTest < Rails::Generators::TestCase
 
         test 'add primary key type with rails 5 when specified in rails generator' do
           run_generator %w[--primary_key_type=uuid --force]
-          if Rails::VERSION::MAJOR >= 5
+          if Rails::VERSION::MAJOR == 5
```
MaicolBen: why this?

brateq: Ok, I've changed the generator code to allow usage of "primary_key_type" not only for Rails 5, but for everything above 5. Is that ok?

PR #1368, opened 2019-12-19
Add case sensitive option required to prevent deprecation warning in …

niciliketo:
…rails 6


theblang:
@niciliketo @MaicolBen FYI, this change actually caused a sequential scan on our users table during email registration, due to the `lower(email)` SQL not utilizing our index on the `users.email` column. I think we should instead move this whole validation to a method, and utilize `resource_class.case_insensitive_keys.include?(:email)` to potentially `downcase` before checking `users.email` with an exact value.

MaicolBen:
@theblang thanks for noticing it, can you send a PR?

PR #1378, opened 2020-01-29
Add paranoid mode

### LINKED ISSUE #1100: Paranoid mode is non existent
While in default configuration here: https://github.com/lynndylanhurley/devise_token_auth/blob/master/config/initializers/devise.rb#L73 it seems that paranoid mode is supported, in fact it isn't at all. I stumbled upon this by accident when I was trying out the password reset api endpoint. It yields different messages depending on whether the email exists in the database or not. With `config.paranoid = true` this behavior isn't changed. I thought it's isolated case, but when I tried to search the repository for `paranoid`, one and only entry was... in that initializer file I mentioned above. Ergo - it's not implemented.

Is this on a roadmap maybe?

 * **Version:** 0.1.42

luisalima:
This PR adds support for Devise paranoid mode and fixes #1100 .
I tried to follow existing conventions from the gem as much as possible, and added a few extra tests for the messages (also when not in paranoid mode).


app/controllers/devise_token_auth/confirmations_controller.rb
```diff
@@ -53,14 +53,23 @@ def render_create_error_missing_email
     end
 
     def render_create_success
+      message = if Devise.paranoid
```
MaicolBen: I don't like this identation, we waste a lot of space, but if you prefer it this way, we can keep it

app/controllers/devise_token_auth/unlocks_controller.rb
```diff
@@ -61,9 +61,14 @@ def render_create_error_missing_email
     end
 
     def render_create_success
+      message = if Devise.paranoid
```
MaicolBen: same, I see this pattern repeated multiple times, maybe we can create a method that for a I18n key, it returns the paranoid version if `Devise.paranoid` is present

app/controllers/devise_token_auth/confirmations_controller.rb
luisalima: @MaicolBen I'm fine with changing it!

app/controllers/devise_token_auth/unlocks_controller.rb
luisalima: @MaicolBen thanks for the suggestion, I will do it.

luisalima:
@MaicolBen added the suggested changes. Thank you!

luisalima:
@MaicolBen the build seems to be failing, however it looks like something on the configuration.

app/controllers/devise_token_auth/passwords_controller.rb
```diff
@@ -181,7 +181,11 @@ def password_resource_params
     end
 
     def render_not_found_error
-      render_error(404, I18n.t('devise_token_auth.passwords.user_not_found', email: @email))
+      if Devise.paranoid
+        render_error(404, I18n.t('devise_token_auth.passwords.sended_paranoid'))
```
ldthorne: Just as you've updated the response message to indicate success regardless of whether or not it worked, I think this should respond with a 200 to be consistent with the success response.

ldthorne: Just as you've updated the response message to indicate success regardless of whether or not it worked, I think this should respond with a 200 to be consistent with the success response.

luisalima: Makes sense, apologies for the delay, will get on it now.

MaicolBen:
@luisalima It's failing for mongo, the codeclimate error configuration was already fixed on master

app/controllers/devise_token_auth/passwords_controller.rb
wilkinsondarolt: @luisalima do you need some help to ship this PR?

MaicolBen:
I will merge it after releasing a patch version with some bugfixes

app/controllers/devise_token_auth/passwords_controller.rb
luisalima: @wilkinsondarolt yes most likely... Can't really take it right now :(

luisalima:
Thanks @MaicolBen and sorry for the delay in the response, the pandemic hit hard here.

phanyzewski:
@MaicolBen is there anything left to be done before getting this merged in?

devaviary:
I noticed that with this commit, you can still distinguish between registered emails and unregistered ones in paranoid mode. Like @ldthorne said, the http status needs to be `200` for both cases, and the response message should not be in `errors`. 

Has this been fixed or is anyone working on this already?

PR #1381, opened 2020-02-03
Serialize updated at without to s

briandunn:
The project I'm on has an overridden ::Time::DATE_FORMATS[:default] that does not include the seconds in the serialized time string. This broke batch detection in a way that was very difficult to track down. This PR uses [iso8601](https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html#method-i-iso8601) To serialize reliably.

### DIFF
--- a/app/models/devise_token_auth/concerns/tokens_serialization.rb
+++ b/app/models/devise_token_auth/concerns/tokens_serialization.rb
@@ -1,18 +1,14 @@
 module DeviseTokenAuth::Concerns::TokensSerialization
+  extend self
   # Serialization hash to json
-  def self.dump(object)
-    object.each_value do |value|
-      if (updated_at_key = ['updated_at', :updated_at].find(&value.method(:[])))
-        if value[updated_at_key].respond_to?(:iso8601)
-          value[updated_at_key] = value[updated_at_key].iso8601
-        end
-      end
-      value.compact!
-    end unless object.nil?
+  def dump(object)
+    JSON.generate(object && object.transform_values do |token|
+      serialize_updated_at(token).compact
+    end.compact)
   end
 
   # Deserialization json to hash
-  def self.load(json)
+  def load(json)
     case json
     when String
       JSON.parse(json)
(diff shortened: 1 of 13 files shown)


briandunn:
@MaicolBen Here is a passing version.

jeremy-ebler-vineti:
`#to_s` is an alias of [to_formatted_s](https://apidock.com/rails/Time/to_formatted_s), which interacts with `Time::DATE_FORMATS` and Rails' I18n. It makes sense to explicitly use a standardized format, like iso8601.

app/models/devise_token_auth/concerns/tokens_serialization.rb
```diff
@@ -1,12 +1,14 @@
 module DeviseTokenAuth::Concerns::TokensSerialization
+  extend self
```
MaicolBen: 👏

app/models/devise_token_auth/concerns/tokens_serialization.rb
```diff
@@ -16,4 +18,16 @@ def self.load(json)
       json
     end
   end
+
+  private
+
+  def serialize_updated_at(token)
+    updated_at_key = ['updated_at', :updated_at].find(&token.method(:[]))
+
+    if token[updated_at_key].respond_to?(:iso8601)
```
MaicolBen: maybe we can write this better:
```ruby
return token unless token[updated_at_key].respond_to?(:iso8601)

token.merge updated_at_key => token[updated_at_key].iso8601

```

test/models/concerns/tokens_serialization_test.rb
```diff
@@ -55,16 +57,48 @@
         assert_equal(ts.dump({}), '{}')
       end
 
-      it 'deserialize tokens' do
```
MaicolBen: why did you remove this spec? are we having breaking changes here?

gemfiles/rails_5_1_mongoid_7.gemfile
```diff
@@ -34,7 +34,7 @@ end
 
 group :test do
   gem "rails-controller-testing"
-  gem "simplecov", require: false
+  gem "simplecov", "< 0.18", require: false
```
MaicolBen: I don't know why this change but it's ok

bdvineti: Newer simplecov is incompatible with codeclimate. https://github.com/codeclimate/test-reporter/issues/413

MaicolBen:
I rebased it in #1423

PR #1386, opened 2020-03-02
Validate that token is valid for patch request last token

ahmedmagdy711:
When in patch requests if any token is passed with right client id it passes and consider the user as logged in because we only that last token which is saved in the db is a valid hash and we don't check it against the token sent by the user.


MaicolBen:
Can you rebase?

PR #1387, opened 2020-03-03
Added token validation to batch requests.

xbits:
Tokens in batch requests where not being validated

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -147,7 +147,7 @@ module DeviseTokenAuth::Concerns::User
       updated_at.to_time > Time.zone.now - DeviseTokenAuth.batch_request_buffer_throttle &&
 
       # ensure that the token is valid
-      DeviseTokenAuth::TokenFactory.valid_token_hash?(last_token)
+      DeviseTokenAuth::Concerns::User.tokens_match?(last_token, token)
     )
   end
 
--- a/lib/devise_token_auth/version.rb
+++ b/lib/devise_token_auth/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module DeviseTokenAuth
-  VERSION = '1.1.4'.freeze
+  VERSION = '1.1.3'.freeze
 end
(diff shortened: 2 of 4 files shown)


lib/devise_token_auth/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module DeviseTokenAuth
-  VERSION = '1.1.3'.freeze
+  VERSION = '1.1.4'.freeze
```
MaicolBen: why this change?

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -147,7 +147,7 @@ def token_can_be_reused?(token, client)
       updated_at.to_time > Time.zone.now - DeviseTokenAuth.batch_request_buffer_throttle &&
 
       # ensure that the token is valid
-      DeviseTokenAuth::TokenFactory.valid_token_hash?(last_token)
+      DeviseTokenAuth::Concerns::User.tokens_match?(last_token, token)
```
MaicolBen: This should be in the factory, it doesn't make any sense to call a user's method for this

lib/devise_token_auth/version.rb
xbits: I guess it doesn't make sense. Not used to making contributions to public repos.

app/models/devise_token_auth/concerns/user.rb
xbits: Same method call is  used right in the previous method, and tokens_match? is defined only on the user.

lib/devise_token_auth/version.rb
MaicolBen:
```suggestion
  VERSION = '1.1.3'.freeze
```

app/models/devise_token_auth/concerns/user.rb
MaicolBen: I think this doesn't make sense anymore as we have this in master https://github.com/lynndylanhurley/devise_token_auth/commit/4a103eddfa503354356704dcd2163e3cfc1a1f16

Issue #1358, opened 2019-11-21
ArgumentError: Argument Out of Range when authorizing via token


TruRooms:
Authentication via the new user session path is working as expected. When a call is made to any controller that calls `authenticate_user!` returns the following error log.

* **Version**: 
Devise Token Auth => 1.1.3, 
Postman => 7.7.3, 
Ruby => 2.6.0, 
Rails => 5.2.3, 
Devise => 4.7.1

* **Request and response headers**: 

*Request Headers*
```
Content-Type: Application/json
access-token: QCsBbXZ96ugarGsuwLPupQ
token-type: Bearer
expiry: 1575572963
uid: noname@test.com
client: 81zmic27JPNh5WNhi9DL3Q
```
*Response Headers*
```
Date: Thu, 21 Nov 2019 19:44:12 GMT
Server: Apache/2.4.38 (Unix) OpenSSL/1.0.2s PHP/7.2.24 Phusion_Passenger/6.0.2
Vary: Origin
X-Runtime: 0.319028
X-Request-Id: bd5d8844-29f9-4a84-b41b-1229b6c71222
X-Powered-By: Phusion Passenger 6.0.2
Content-Length: 110718
Status: 500 Internal Server Error
Connection: close
Content-Type: text/html; charset=UTF-8
```

* **Rails Stacktrace**: 
```
Processing by Api::HomesController#index as JSON
  Parameters: {"subdomain"=>"api", "home"=>{}}
  User Load (0.7ms)  SELECT  "users".* FROM "users" WHERE "users"."uid" = $1 LIMIT $2  [["uid", "noname@test.com"], ["LIMIT", 1]]
  ↳ /usr/local/Cellar/passenger/6.0.2_1/libexec/src/ruby_supportlib/phusion_passenger/rack/thread_handler_extension.rb:97
  Rendering api/homes/index.json.jbuilder
  Home Load (0.4ms)  SELECT "homes".* FROM "homes"
  ↳ app/views/api/homes/index.json.jbuilder:1
  Rendered api/homes/index.json.jbuilder (6.7ms)
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ /usr/local/Cellar/passenger/6.0.2_1/libexec/src/ruby_supportlib/phusion_passenger/rack/thread_handler_extension.rb:97
   (0.1ms)  BEGIN
[...]

PR #1388, opened 2020-03-07
Issue - 1358 Argument error when converting token updated_at using to…

saichander17:
Fix for #1358 .

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -158,7 +158,7 @@ module DeviseTokenAuth::Concerns::User
     token = create_token(
       client: client,
       last_token: tokens.fetch(client, {})['token'],
-      updated_at: now
+      updated_at: now.to_s(:rfc822)
     )
 
     update_auth_header(token.token, token.client)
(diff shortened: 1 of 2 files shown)


app/models/devise_token_auth/concerns/user.rb
```diff
@@ -158,7 +158,7 @@ def create_new_auth_token(client = nil)
     token = create_token(
       client: client,
       last_token: tokens.fetch(client, {})['token'],
-      updated_at: now
+      updated_at: now.to_s(:rfc822)
```
MaicolBen: wouldn't this break other's people defaults?

saichander17: @MaicolBen , Should it not be used like updated_at.to_time. Even in the whole code of this particular gem, whenever it needs to be compared with time, to_time was always used.

MaicolBen:
I will rollback this change, it breaks #1423 which was requested by a lot of people (compared with this one that is only you)

MaicolBen:
If you want this change back, please add a new PR with a test that verifies the bug. Maybe you have to read `Time::DATE_FORMATS[:default]`

PR #1407, opened 2020-06-05
Ignore sync uid in case confirmable mail changed

pnghai:


iMacTia:
@MaicolBen @pnghai I'm using the `master` branch in my project and this PR has broken all tests.
Even though I didn't setup omniauth, the module is still included [here](https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L33) and now all users I create providing only email and password have an empty `uid`. So only the first user gets saved, all others trigger a "ERROR:  duplicate key value violates unique constraint "index_users_on_uid_and_provider"

MaicolBen:
That concern is for everyone, not only omniauth. Did you override `provider`

iMacTia:
@MaicolBen no I didn't, it's still the default value (`email`). I reverted the implementation of this method and everything works as expected, but with this new implementation it doesn't set the `uid` automatically as it did before.
I believe my issue is that I've added the `confirmable` module and the code is going into the if branch and returning without setting the `uid`

PR #1420, opened 2020-07-07
Allow customized controllers to be added to devise_for

nnluukhtn:
# Why?

* Currently there is no way to add customized controller when using `mount_devise_token_auth_for`. See more at https://github.com/lynndylanhurley/devise_token_auth/blob/master/lib/devise_token_auth/rails/routes.rb#L10-L23. For eg: adding customized controller for https://github.com/scambra/devise_invitable#configuring-controllers-.

# How?

* Allow controller option when using `mount_devise_token_auth_for`. Eg:
```ruby
  mount_devise_token_auth_for 'User', at: 'users', skip: [], controllers: {
      sessions:  'api/users/sessions',
      passwords:  'api/users/passwords',
      token_validations:  'api/users/token_validations',
      invitations: 'api/users/invitations'
    }
```

### DIFF
--- a/lib/devise_token_auth/rails/routes.rb
+++ b/lib/devise_token_auth/rails/routes.rb
@@ -17,10 +17,12 @@ module ActionDispatch::Routing
       unlocks_ctrl           = opts[:controllers][:unlocks] || 'devise_token_auth/unlocks'
 
       # define devise controller mappings
-      controllers = { sessions: sessions_ctrl,
-                      registrations: registrations_ctrl,
-                      passwords: passwords_ctrl,
-                      confirmations: confirmations_ctrl }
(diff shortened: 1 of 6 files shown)


MaicolBen:
It seems that is breaking the specs for some reason, can you take a look? I don't know if it would be easier to add a new spec for it

nnluukhtn:
Agree, let me check and add more specs for this @MaicolBen

nnluukhtn:
@MaicolBen Updated with new specs and fix. Please check whether it's good to merge or not

MaicolBen:
Yes, but still some specs are failing

MaicolBen:
#1435 was merged and has these changes

PR #1431, opened 2020-10-05
Enhancement: remove unnecessary statement in destroy session

martinjaimem:
Small code enhancement, removes an unnecessary check for the presence of the `client`.

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -48,7 +48,7 @@ module DeviseTokenAuth
     def destroy
       # remove auth instance variables so that after_action does not run
       user = remove_instance_variable(:@resource) if @resource
-      client = @token.client if @token.client
+      client = @token.client
       @token.clear!
 
       if user && client && user.tokens[client]


app/controllers/devise_token_auth/sessions_controller.rb
```diff
@@ -48,7 +48,7 @@ def create
     def destroy
       # remove auth instance variables so that after_action does not run
       user = remove_instance_variable(:@resource) if @resource
-      client = @token.client if @token.client
+      client = @token.client
```
MaicolBen, line 51: what does this line really do?

martinjaimem: just sets the client, there is no need to set it as `client = @token.client if @token.client` the `if` is unnecessary. So instead of doing:
```
client = @token.client if @token.client
```

We could just be doing:

```
client = @token.client
```

I know it is something tiny but in the end, removing the `if` helps with the readability and complexity.

MaicolBen (approved):

app/controllers/devise_token_auth/sessions_controller.rb
MaicolBen: yeah but client is an unused variable

martinjaimem: I think it is being used in line 55:
```
user.tokens.delete(client)
```

martinjaimem:
@MaicolBen Is there anything that is missing on my end in order to merge this?

app/controllers/devise_token_auth/sessions_controller.rb
MaicolBen: cool

PR #1436, opened 2020-11-02
Multi token lifespan

aleochoam:
When using multiple User classes, the token lifespan is the same for every class, this PR allows to override this default token and use a custom one in a specific class

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -90,6 +90,7 @@ module DeviseTokenAuth::Concerns::User
     end
 
     def create_token(client: nil, lifespan: nil, cost: nil, **token_extras)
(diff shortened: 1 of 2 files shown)


app/models/devise_token_auth/concerns/user.rb
```diff
@@ -254,4 +255,8 @@ def clean_old_tokens
       tokens.shift while max_client_tokens_exceeded?
     end
   end
+
+  def token_lifespan
+    DeviseTokenAuth.token_lifespan
```
MaicolBen, line 260: now, what's the purpose of this code https://github.com/lynndylanhurley/devise_token_auth/blob/ce2cf2c180e6632dd054dd4b099e27f5a264de83/lib/devise_token_auth/token_factory.rb#L63 ?

PR #1440, opened 2020-11-17
not work alongside standard Devise for class with namespace, use mapping directly as scope

islue:
When using alongside standard Devise, existing user can't be found in case of class with namespace.

https://github.com/heartcombo/devise/blob/master/lib/devise/controllers/helpers.rb#L48
`rc` would be something like `myapp/user`, use `mapping` directly as `scope` to keep consistency of behavior with devise.


MaicolBen (approved):
Can you repush or something? I think the build failed randomly

islue:
repushed without code changes

PR #1447, opened 2020-12-10
Fix critical error on registration with confirmation mode

pnghai:
In case confirmation is activated:
- When more than 1 user registered, the 'uid' column will be left empty until confirmation complete. That will lead to duplication of 'uid-provider' index as uid are both nulled but providers are the same 'email'


MaicolBen:
Can you add a test confirming the bug?

Hoss3770:
The problem appeared again in version 1.1.4 with rails 6.1

PR #1448, opened 2020-12-17
Enhancement: render auth headers only in the response body on sign in

martinjaimem:
### Description:

In this PR there is an enhancement for the session's creation. On sign-in, the client [should receive the "auth headers" in the body](https://tools.ietf.org/html/rfc6750#page-10).

Specifically, according to the IETF documentation:
```
5.1.  Successful Response

   The authorization server issues an access token and optional refresh
   token, and constructs the response by adding the following parameters
   to the entity-body of the HTTP response with a 200 (OK) status code:

   access_token
         REQUIRED.  The access token issued by the authorization server.

   token_type
         REQUIRED.  The type of the token issued as described in
         Section 7.1.  Value is case insensitive.

   expires_in
         RECOMMENDED.  The lifetime in seconds of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value.
```

[Here is the link to it.](https://tools.ietf.org/html/rfc6749#section-5.1)

This is a breaking change for all the clients using this gem, so I'd like some input on how to improve this and add some warnings to the users 😄

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -6,6 +6,8 @@ module DeviseTokenAuth
     before_action :set_user_by_token, only: [:destroy]
     after_action :reset_session, only: [:destroy]
 
+    skip_after_action :update_auth_header, only: [:create]
+
     def new
       render_new_error
     end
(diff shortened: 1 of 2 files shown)


MaicolBen:
What does this do for you? And why we need to follow OAuth 2 RFC? AFAIK We don't follow OAuth 2 except in the Omniauth part

pablanco:
> What does this do for you? And why we need to follow OAuth 2 RFC? AFAIK We don't follow OAuth 2 except in the Omniauth part

You should use `OAuth 2 RFC` because it's a standard, shouldn't you?

MaicolBen:
But can you explain to me why our auth system is OAuth? We use email/password authentication.

PR #1449, opened 2020-12-25
Fix, Avoid checking resources multi times

leanhthang:
When use devise_token_auth_group the method 'set_user_by_token' has been call multi times per mapping name. so that, resource has found not reused. That will Increase performance.


MaicolBen:
Can you fix the specs?

leanhthang:
> Can you fix the specs?

Can you help me explain for this case test why used refute_equal? 
```ruby
    it 'should not define current_mang' do
      refute_equal @resource, @controller.current_mang
    end
```

leanhthang:
> Can you fix the specs?
My approach is to assign the current user default if it exists, So, I'm clear all case tests that should not define current_user.

PR #1450, opened 2021-01-03
Fix Rails 6.1 Deprecation warning by updating connection_config to connection_db_config

csalvato:
Rails 6.1 generates a deprecation warning since connection_config will be removed from Rails 6.2.

This resolves that issue by changing the symbol where it's referenced.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -20,7 +20,7 @@ module DeviseTokenAuth::Concerns::ResourceFinder
   end
 
   def find_resource(field, value)
-    @resource = if resource_class.try(:connection_config).try(:[], :adapter).try(:include?, 'mysql')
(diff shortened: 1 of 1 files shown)


MaicolBen (approved):
Weird that tests fail, so I cannot merge it

csalvato:
> Weird that tests fail, so I cannot merge it

Yes and I’m not sure how to resolve those failures 😞

csalvato:
Does anyone know why mysql would be failing on this PR?

csalvato:
My guess is that this object no longer represents 'mysql' in the same way in the adapter.

gatperdut:
will this be merged? I do get the deprecation warning as well.

MaicolBen:
Needs to be rebased, also I wonder if removing this break something

MaicolBen:
I merged https://github.com/lynndylanhurley/devise_token_auth/pull/1467, so probably we don't need this

PR #1452, opened 2021-01-05
Fix, Avoid checking resources multi times

leanhthang:
When use devise_token_auth_group the method 'set_user_by_token' has been call multi times per mapping name. so that, resource has found not reused. That will Increase performance.


app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -159,4 +163,10 @@ def auth_header_from_batch_request
     end
     auth_header
   end
+
+  def resource_is_checked?(mapping)
+    return true if @resource || eval("@has_checked_#{mapping}") == true
```
MaicolBen, line 168: Ugh, this is ugly (the `eval`), why do we need to have another instance variable?

MaicolBen (commented):
Thanks for adding the tests

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
leanhthang: I think that is useful, When i have a many  `mapping` and not yet sign in, this instance variable to reject no need many DB connections.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -24,6 +24,9 @@ def set_request_start
 
   # user auth
   def set_user_by_token(mapping = nil)
+    # Avoid checking resources multi times
+    return @resource if resource_is_checked?(mapping)
```
booleanbetrayal, line 28: should we just use `defined?(@resource)` instead of this elaborate check?

MaicolBen: Agreed

leanhthang: I'm try set `return @resource if defined?(@resource)`, It works fine when logged in. 
But, when not logged in it still checks resources many times.
Screenshot from 2021-01-28 10-49-23

Issue #1371, opened 2020-01-11
Does DTA support HTTP Only Cookie Refresh tokens along site access tokens?


aarona:
Also, does it have JWT support? I'd like to have silent refresh support.

theblang:
@aarona As of #1453 you can now use HttpOnly cookies for transferring the token. In #1281 I've also brought up adding the ability to turn off the token transfer via headers entirely for the security aspect.

MaicolBen:
>does it have JWT support? 

It doesn't, feel free to add it. But I'd recommend just going straight to `jwt` and implement auth with that

Issue #1321, opened 2019-07-29
Storing auth token in browser best practise


sebmor:
With many people using rails as a standalone API with a Vue/React/... frontend. I ended up storing the token as a cookie.

One thing I am concerned about is that it saves the email as plaintext in the cookie, even though the added vector of attack on this is relatively small, it makes me feel uncomfortable when user privacy is important for your application.

I wanted to hear from other people in my usecase and if it's perhaps worth investigating if this library potentially should allow to change your uid field or encrypt the uid by default?

MaicolBen:
Closing as duplicate of #1281, there is no best answer but it might help you.

theblang:
@MaicolBen Happened to notice, this ticket can be closed in favor of discussion in #1281

Issue #1281, opened 2019-04-17
Where to store token securely?


Uysim:
I have been review this https://github.com/lynndylanhurley/devise_token_auth/issues/1005 the answer in this issue is not correct. We need to guide the way to handle token securely on client side. Nothing I know from this gem to store token securely. 

I think the better way is to handle token with HttpOnly cookie

MaicolBen:
I don't have experience with HttpOnly cookie, but can you submit a PR suggesting the best security recommendation according to you?

CDimonaco:
This gem provide the secure token, how you want store it, it's a completely different stuff, in a single page application context is correct to store the token on the local/session storage, according to your needs, you can also put in the cookie, but it's just the same level of security, a cookie can be also stolen on the client side. If you want more security you can just encrypt your local storage data, but it's a false sense of security, because you will need the key for decrypt the stuff, and it's also on the client side.

If you want stuck with the token in a cookie, I suggest you to use the standard Devise gem. using a cookie for the authentication facilities that this gem offers, does not makes sense, just imagine the process of change the authentication token and other 'headers' stuff, of this gem, and at each request recreating the cookie, it's just an overkill and overhead.

Uysim:
@MaicolBen I also never experience HttpOnly cookie as I use this gem for years. 

But recently I research deep on this on the internet, I can see HttpOnly cookie is the cookie that can only read on the server. So no one on client side can access it. Every request browser will attach this kind of cookie to the server automatically.

I try to read docs and review the source. Seem like it no way to have this kind of security in this gem yet. Also, currently I try to implement it by myself. But it need ton of customization. 

That why I come to ask the suggestion from community. If anyone use to take care of this security by simplify?

Uysim:
@CDimonaco Actually, it is not completely different staff. It is the same the way you provide a secure token and there is no way to store that secure token by the way you provide. 

If you work with Mobile application, it is ok to store that kind of token on the client side. But if you working with single page application it is the scary part. 

1. Do you feel safe it under local storage? Some can get by just run `localStorage.getItem('access_token')` on inspect element.

2. Session storage maybe more secure but you client will need to login with username and password all the time after they closed their browser

3. Devise gem is good, but I think it far beyond the one page application. 

4. Would you scarify the security performance over the performance? We should won't our user data at risk because of the performance. 

I think we should concern the way provide the secure token in order be more secure. What is your idea? Let's me know

[...]

PR #1453, opened 2021-01-05
Add support for sending and receiving the auth token via a server cookie

theblang:
There are two main reasons to use a server cookie to send and receive the auth token.

1) For sharing auth state across subdomains with the `Domain` attribute. From [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies):

> The Domain attribute specifies which hosts are allowed to receive the cookie. If unspecified, it defaults to the same origin that set the cookie, excluding subdomains. If Domain is specified, then subdomains are always included. Therefore, specifying Domain is less restrictive than omitting it. However, it can be helpful when subdomains need to share information about a user.

Note that you **cannot** access `localStorage` across subdomains.

2) For security reasons. If you only send the auth token in an `HttpOnly` server cookie then JavaScript can never access it, which helps protect against XSS attacks. These three issues are mostly pertaining to the security aspect: https://github.com/lynndylanhurley/devise_token_auth/issues/1281 https://github.com/lynndylanhurley/devise_token_auth/issues/1321 https://github.com/lynndylanhurley/devise_token_auth/issues/1371

This PR introduces a server cookie as an additive opt-in feature to send and receive the auth token, **which will allow for the subdomain sharing mentioned in point one**. By introducing the option to use a server cookie, we lay the groundwork for addressing the security aspect in point two. **I didn't want to make the changes for point two yet** as that will be more impactful, because we'd need to make it possible to **not** expose the auth token in headers. I imagine if and when we do that work we'll probably add an analogous `headers_enabled` config to hinge off of in code to turn off sending and receiving the auth token in headers.

cc @booleanbetrayal @nbrustein @apuntovanini


app/controllers/devise_token_auth/sessions_controller.rb
```diff
@@ -55,6 +55,16 @@ def destroy
         user.tokens.delete(client)
         user.save!
 
+        if DeviseTokenAuth.cookie_enabled
+          if DeviseTokenAuth.cookie_attributes[:domain]
```
theblang, line 59: I just realized that we might need to be smarter about the `domain` attribute. We might need to determine it by looking at the request's domain (in which case we'd probably want to allow configuring a whitelist, like we do with `redirect_whitelist`). This is because the `Domain` cookie attribute only allows one value, so as the code sits now you wouldn't be able to handle multiple domains being served from the same box.

apuntovanini:
I tested in our application and works like charm. I only have issues logging out (we use a graphql endpoint to make our calls, we should probably switch to classic controllers flow for authentication instead). But sharing cookies with our subdomains is crucial to us, well done!

app/controllers/devise_token_auth/sessions_controller.rb
apuntovanini: Make sense, remember to use allowlist instead of whitelist

theblang: @apuntovanini Another option is to just support passing to the `cookie_attributes[:domain]` config either a string, which will work for most people, or a proc, which could accept the request and do whatever logic (in our case, return the domain of the request so long as it matches the allow list).

theblang:
> I tested in our application and works like charm. I only have issues logging out (we use a graphql endpoint to make our calls, we should probably switch to classic controllers flow for authentication instead). But sharing cookies with our subdomains is crucial to us, well done!

Yay, thanks so much! I'm so glad this will help y'all as well!

app/controllers/devise_token_auth/sessions_controller.rb
theblang: [Here](https://github.com/lynndylanhurley/devise_token_auth/pull/1453/commits/7592336caa29f8a1c4d9173f409993becba6e1ef) are the changes for that second strategy. And my config looks like:

```
    config.cookie_attributes = {
        domain: lambda { |request| request.domain if valid_cookie_domains.include?(request.domain) },
        secure: Rails.env.production?,
        httponly: true,
        same_site: 'Strict',
        expires: token_lifespan.from_now,
        encrypt: true
    }
```

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -113,6 +126,18 @@ def update_auth_header
...
+    # Most people will just set a string for the domain attribute config. But if you need
+    # more flexibility, such as to dynamically choose the domain when serving multiple domains
+    # from a single server, you can set a Proc that will be passed the request.
+    config_value = DeviseTokenAuth.cookie_attributes[:domain]
+    if config_value.is_a?(String)
+      config_value
+    elsif config_value.is_a?(Proc)
+      config_value.call(request)
+    end
+  end
```
theblang, line 139: See [this discussion](https://github.com/lynndylanhurley/devise_token_auth/pull/1453#discussion_r552158961)

.gitignore
```diff
@@ -9,6 +9,7 @@ test/dummy/.sass-cache
 test/dummy/config/application.yml
 coverage
 .idea
+.byebug_history
```
theblang, line 12: Drive-by that popped up while debugging a test

app/controllers/devise_token_auth/sessions_controller.rb
```diff
@@ -55,6 +55,17 @@ def destroy
         user.tokens.delete(client)
         user.save!
 
+        if DeviseTokenAuth.cookie_enabled
+          cookie_domain = get_cookie_domain
+          if cookie_domain.present?
+            # If a cookie is set with a domain specified then it must be deleted with that domain specified
+            # See https://stackoverflow.com/a/6244724/1747491
+            cookies.delete(DeviseTokenAuth.cookie_name, domain: cookie_domain)
```
MaicolBen, line 63: This line is unreached by specs but it could be ok

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -123,11 +148,30 @@ def refresh_headers
...
     end # end lock
   end
 
+  def set_cookie(auth_header)
+    cookie_attributes = DeviseTokenAuth.cookie_attributes
+    cookie_attributes.merge!(value: auth_header.to_json)
+
+    cookie_domain = get_cookie_domain
+    if cookie_domain.present?
+      cookie_attributes.merge!(domain: cookie_domain)
```
MaicolBen, line 169: This line is unreached by specs but it could be ok

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -113,6 +126,18 @@ def update_auth_header
     end
   end
 
+  def get_cookie_domain
+    # Most people will just set a string for the domain attribute config. But if you need
+    # more flexibility, such as to dynamically choose the domain when serving multiple domains
+    # from a single server, you can set a Proc that will be passed the request.
+    config_value = DeviseTokenAuth.cookie_attributes[:domain]
+    if config_value.is_a?(String)
```
MaicolBen, line 134: These cases are unreached in specs but that could be ok

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
```diff
@@ -35,11 +35,20 @@ def set_user_by_token(mapping = nil)
     access_token_name = DeviseTokenAuth.headers_names[:'access-token']
     client_name = DeviseTokenAuth.headers_names[:'client']
 
+    # gets values from cookie if configured and present
```
MaicolBen, line 38: can all this new cookie code be in a new module like `set_user_by_cookie` or something like that, so we don't pollute more this big file, you can include the concern/module in this file

MaicolBen (commented):
I don't like using cookies in APIs but your reasoning of why is a good alternative sounds good

apuntovanini:
> I don't like using cookies in APIs but your reasoning of why is a good alternative sounds good

Just to understand, how would how tackle cross-subdomain authentication with api without a cookie? Thanks!

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
theblang: This conditional will be moot now, see [this recent commit](https://github.com/lynndylanhurley/devise_token_auth/pull/1453/commits/536340b3e1861fef0e595158b3588b747d765898).

app/controllers/devise_token_auth/sessions_controller.rb
theblang: This conditional will be moot now, see [this recent commit](https://github.com/lynndylanhurley/devise_token_auth/pull/1453/commits/536340b3e1861fef0e595158b3588b747d765898).

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
theblang: > I don't like using cookies in APIs but your reasoning of why is a good alternative sounds good

Yeah, we weren't too keen on it either. But it's the only way to share state across subdomains. And the general consensus seems to be that it's more secure for auth (once you're using **only** the HttpOnly cookie, which we'll hopefully allow in later work).

theblang:
> I don't like using cookies in APIs but your reasoning of why is a good alternative sounds good

Yeah, we weren't too keen on it either. But it's the only way to share state across subdomains. And the general consensus seems to be that it's more secure for auth (once you're using **only** the HttpOnly cookie, which we'll hopefully allow in later work).

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
theblang: I can, my only push back would be that hopefully in the future we'll allow turning off headers (for the security reasons mentioned in point two of the description). In that case, it might feel a bit disjointed to have the header logic in `set_user_by_token` while the cookie logic is in `set_user_by_cookie`, when really both logics deal with handling the token. In that future world, would we move the header logic to another concern called `set_user_by_headers`? If so, wouldn't `set_user_by_token` just become a skeleton file?

app/controllers/devise_token_auth/sessions_controller.rb
theblang: 🤦  I just saw this in the [in the Rails docs](https://api.rubyonrails.org/classes/ActionDispatch/Cookies.html):

> To support multiple domains, provide an array, and the first domain matching request.host will be used. Make sure to specify the :domain option with :all or Array again when deleting cookies.

So I can totally gut this Proc logic now, and instead just pass along either a string or Array from the config to the cookie call.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
theblang: On this note, see [my recent realization](https://github.com/lynndylanhurley/devise_token_auth/pull/1453#discussion_r553446626). That will clean up the amount of noise in this file considerably, commit incoming.

app/controllers/devise_token_auth/sessions_controller.rb
theblang: Actually, using an array doesn't quite work for us since the `request.host` that it matches against includes the subdomain. But I think the `:all` option will:

> If you use a schema like www.example.com and want to share session with user.example.com set :domain to :all

theblang: See [this commit](https://github.com/lynndylanhurley/devise_token_auth/pull/1453/commits/66f8fa98cfbefbeaa5f098f3efa482744f9e668d). So much cleaner! I wish I had realized that earlier, lol.

app/controllers/devise_token_auth/concerns/set_user_by_token.rb
theblang: Moot now that I've gutted the unnecessary Proc logic.

theblang:
@MaicolBen @booleanbetrayal Could one of you retry the failed Travis job? It seems that just one of the jobs failed with:

```
Fetching nokogiri 1.10.10
Bundler::GemspecError: Could not read gem at
/home/travis/build/lynndylanhurley/devise_token_auth/gemfiles/vendor/bundle/ruby/2.3.0/cache/nokogiri-1.10.10.gem.
It may be corrupted.
An error occurred while installing nokogiri (1.10.10), and Bundler
cannot continue.
Make sure that `gem install nokogiri -v '1.10.10' --source
'https://rubygems.org/'` succeeds before bundling.
In rails_5_1.gemfile:
  devise_token_auth was resolved to 1.1.4, which depends on
    rails was resolved to 5.2.4.4, which depends on
      actioncable was resolved to 5.2.4.4, which depends on
        actionpack was resolved to 5.2.4.4, which depends on
          actionview was resolved to 5.2.4.4, which depends on
            rails-dom-testing was resolved to 2.0.3, which depends on
              nokogiri
```

booleanbetrayal:
@theblang - running this now produces a new error: 

`You are using an old OmniAuth version, please ensure you have 1.0.0.pr2 version or later installed.`

It looks like Devise has [hard-coded the omniauth dependency](https://github.com/heartcombo/devise/issues/5326) in a way that it's breaking it's 2.0.0 release. It would be preferable if they fixed this on their end, but we might want to constrain the Gem version in the project if they are slow with a release. (cc: @MaicolBen )

MaicolBen:
I'm going to wait for https://github.com/heartcombo/devise/issues/5326 for a few days

theblang:
[Looks like](https://github.com/heartcombo/devise/pull/5327#issuecomment-769811843) we're close!

MaicolBen:
Can you rebase/update with last master?

PR #1463, opened 2021-02-09
Fix omniauth version until devise fixes omniauth requirement

MaicolBen:


MaicolBen:
I will fix the mongo error in a separate PR, so we can merge other PRs

PR #1467, opened 2021-02-25
Update connection_config to connection_db_config

melnik0v:
Rails 6.1 generates a deprecation warning since connection_config will be removed from Rails 6.2.

### DIFF
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -20,12 +20,25 @@ module DeviseTokenAuth::Concerns::ResourceFinder
   end
 
   def find_resource(field, value)
-    @resource = if resource_class.try(:connection_config).try(:[], :adapter).try(:include?, 'mysql')
+    @resource = if database_adapter&.include?('mysql')
                   # fix for mysql default case insensitivity
                   resource_class.where("BINARY #{field} = ? AND provider= ?", value, provider).first
                 else
                   resource_class.dta_find_by(field => value, 'provider' => provider)
                 end
+  end
+
+  def database_adapter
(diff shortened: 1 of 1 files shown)


app/controllers/devise_token_auth/concerns/resource_finder.rb
```diff
@@ -20,7 +20,9 @@ def get_case_insensitive_field_from_resource_params(field)
   end
 
   def find_resource(field, value)
-    @resource = if resource_class.try(:connection_config).try(:[], :adapter).try(:include?, 'mysql')
+    adapter = resource_class.try(:connection_db_config)&.configuration_hash&.try(:[], :adapter)
```
gatperdut, line 23: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connection_config

Since it's no longer a hash but an object, wouldn't it be enough to transform that line into something like

`adapter = resource_class.try(:connection_db_config)&.try(:adapter)`

MaicolBen:
Can you rebase? So I can see the test pass

melnik0v-me:
@MaicolBen @gatperdut rebased and updated

melnik0v-me:
seems like failed specs are not related to my changes, something wrong with coverage tool

MaicolBen (approved):

gatperdut:
will there be a version bump for this?

experimatt:
Hi, same question as above. Anything we can do to help get a release out before Rails 6.2?

s-aida:
Meanwhile I have to suppress the warning in this way.
```ruby
class ApplicationRecord < ActiveRecord::Base
  self.abstract_class = true

  def self.connection_config
    connection_db_config
  end
end

```

PR #1468, opened 2021-02-27
Replace URI::escape which was removed in Ruby 3

alea12:
`URI::escape` was removed in Ruby 3.0.0. https://github.com/ruby/uri/pull/9
This PR replaces `URI::escape` with `CGI::escape`, making the library compatible with Ruby 3.


MaicolBen:
`CGI` doesn't work as `URI`, I realized after I merged (the suite was broken so I didn't know), I made #1470 to fix it

Issue #1421, opened 2020-07-20
Trying to integrate with devise-multi_email


mgmodell:
devise_token_auth seems to ignore the emails provided by devise-multi_email. As you can see, it seems to error out in sessions_controller#create when it doesn't find the email field. In this case, I could probably hack the `resource_finder` concern and hardcode the emails table in, but I'd like to do it the right way - except I don't know either package well enough to do so and I'm not sure if that'd be sufficient anyway.

If you can guide me a bit, I can probably put together a PR for this. Am I on the right track? Where should I be expecting to make these changes?

* **Version**: 
- devise_token_auth: 1.1.4
- devise: 4.7.2
- devise-multi_email: 3.0.0

* **Request and response headers**: 
Request URL: http://localhost:3000/api/v1/auth/sign_in
Request Method: POST
Status Code: 500 Internal Server Error
Remote Address: [::1]:3000
Referrer Policy: strict-origin-when-cross-origin
Content-Length: 1477
Content-Type: text/html; charset=UTF-8
X-Request-Id: 8c652be9-3dce-44dc-a996-189979a94f0c
X-Runtime: 0.530246
Accept: */*
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9,ko;q=0.8,he;q=0.7
Connection: keep-alive
Content-Length: 27
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
Cookie: 
_session_id=834a0a4e9a7243567f916f63dba41b90
Host: localhost:3000
Origin: http://localhost:3000
Referer: http://localhost:3000/
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: same-origin
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36
x-kopernio-plugin-version: 0.12.18
X-Requested-With: XMLHttpRequest
email: dfgs
password: sdgfsfd

* **Rails Stacktrace**:
Started POST "/api/v1/auth/sign_in" for ::1 at 2020-07-20 00:25:12 -0400
Processing by DeviseTokenAuth::SessionsController#create as */*
  Parameters: {"email"=>"dfgs", "password"=>"[FILTERED]"}
[...]

PR #1469, opened 2021-03-05
Fix for devise_multi_email support (redux)

mgmodell:
I've cleaned up the code but I need help building the tests. How do I dynamically load/unload the gem and the initializer at test time?

#1421


MaicolBen:
Closing as per my comment on #1421, the code needs improvements but it's useful for future people 🙂

PR #1472, opened 2021-03-10
Workaround for cc-test-reporter with SimpleCov 0.18

enomotodev:
This PR fixes the build error when using cc-test-reporter with SimpleCov 0.18.

This patch is a workaround until the following issue will be resolved.
codeclimate/test-reporter#418


MaicolBen (approved):
Nice!

PR #1478, opened 2021-04-24
fix mongoid detecting bug

qiuyin:
## issue
This will always return string "expression" even when I don't use Mongoid.
```ruby
defined?('Mongoid')
```

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -218,13 +218,8 @@ module DeviseTokenAuth::Concerns::User
   end
 
   def should_remove_tokens_after_password_reset?
-    if Rails::VERSION::MAJOR <= 5 ||defined?(Mongoid)
-      encrypted_password_changed? &&
+    encrypted_password_changed? &&
         DeviseTokenAuth.remove_tokens_after_password_reset
-    else
-      saved_change_to_attribute?(:encrypted_password) &&
-        DeviseTokenAuth.remove_tokens_after_password_reset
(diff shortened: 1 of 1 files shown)


MaicolBen (approved):

qiuyin:
@MaicolBen Thanks for your approving. 
I am newbie for this gem. Could you tell me how to run the test local and why the test is failing now

MaicolBen:
@qiuyin I believe this was a false positive, I don't know how to rerun it, can you rebase/push again?

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -218,13 +218,8 @@ def destroy_expired_tokens
   end
 
   def should_remove_tokens_after_password_reset?
-    if Rails::VERSION::MAJOR <= 5 ||defined?('Mongoid')
-      encrypted_password_changed? &&
+    encrypted_password_changed? &&
         DeviseTokenAuth.remove_tokens_after_password_reset
-    else
-      saved_change_to_attribute?(:encrypted_password) &&
```
qiuyin, line 225: `saved_change_to_attribute?`
- will return false in `before_save` callback
- will return true in `after_save` callback

But `encrypted_password_changed?` is opposite to `saved_change_to_attribute?`
So we must use `encrypted_password_changed?`

qiuyin:
@MaicolBen I have fixed the failed test, Could you review again.
https://github.com/lynndylanhurley/devise_token_auth/pull/1478#discussion_r628898473

Issue #1477, opened 2021-04-24
null user attributes on email registration


kerolloz:
When a user registers using email, the values sent in JSON are not saved in the user object.
![image](https://user-images.githubusercontent.com/36763164/115957280-a9f6b480-a501-11eb-93eb-02421f0e4b8b.png)

PR #1481, opened 2021-04-29
Add new param FAQ

muratiger:
After looking at the following issue, I thought it would be better if the documentation included a way to allow additional parameters when creating or updating an account.

https://github.com/lynndylanhurley/devise_token_auth/issues/1477

If there are any problems with the English or if the explanations are difficult to understand, please feel free to point them out to me.


MaicolBen (approved):
Thanks

PR #1486, opened 2021-05-13
check password changed only when using password authentication

qiuyin:
## what
To support the case that we don't use password authentication (for example, only use omniauth)


MaicolBen (approved):
It'd be great a test but I think the change is pretty simple and harmless

PR #1498, opened 2021-08-05
wrap creation and save of token in a transaction

### LINKED ISSUE #1497: Concurrency issue?
I think there is a potential concurrency issue when creating tokens for the same user.

The tokens are stored in a hash in a JSON column on the User model (or whatever model you have).
The key is the client, value is another Hash containing expiry, the token and extras  
https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/models/devise_token_auth/concerns/user.rb#L92

Loading of the resource, creation of the token and saving ther resource happens here https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/sessions_controller.rb#L21-L30

There is no lock on the resource and it could happen that multiple requests run at "the same time" thus overwriting tokens.

At least this seems like a possible explanation for problems we see in an app that uses background jobs to call an API that uses DeviseTokenAuth.

pascalbetz:
this will reload the record and lock the row in the DB so we
only ever have one request accessing the record at one time.

this prevents multiple requests overwriting the tokens.

We tested this in our app with 10 threads accessing at the "same time" and could show the problem and that it disappears with the fix.
if you have any idea how i can do this with a test, please advise.

Also this will most likely not work with MongoId?

fixes #1497

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -26,10 +26,9 @@ module DeviseTokenAuth
         if (@resource.respond_to?(:valid_for_authentication?) && !@resource.valid_for_authentication? { valid_password }) || !valid_password
           return render_create_error_bad_credentials
         end
-        @resource.with_lock do
-          @token = @resource.create_token
-          @resource.save
-        end
+
+        create_and_assign_token
+
         sign_in(:user, @resource, store: false, bypass: false)
 
         yield @resource if block_given?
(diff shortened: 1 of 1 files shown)


MaicolBen:
You might need to ask if it isn't Mongo so you can use `with_lock`

jvanniew:
I'm seeing a similar issue when tokens are updated on each request.

pascalbetz:
Can you try if my changes fix this for you? 
I'm no longer using devise token auth so getting this merged has not super high prio to me but I'm still ready to change/adapt the merge request if there is a chance to get it merged.

MaicolBen (approved):
Good refactor, thanks

PR #1502, opened 2021-09-02
Fix callback if migrations fails

thooams:
To  prevent failed migration, when you insert this gem in existing project and you want to run: 
```
rake db:create && rake db:migrate
```
Migrations fail because of uid and provider attributes that do not exist yet in the first migrations (past migrations).

### DIFF
--- a/app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
+++ b/app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
@@ -4,12 +4,12 @@ module DeviseTokenAuth::Concerns::UserOmniauthCallbacks
   extend ActiveSupport::Concern
 
   included do
-    validates :email, presence: true,if: :email_provider?
-    validates :email, :devise_token_auth_email => true, allow_nil: true, allow_blank: true, if: :email_provider?
-    validates_presence_of :uid, unless: :email_provider?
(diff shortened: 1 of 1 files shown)


app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
```diff
@@ -26,6 +30,6 @@ def sync_uid
     unless self.new_record?
       return if devise_modules.include?(:confirmable) && !@bypass_confirmation_postpone && postpone_email_change?
     end
-    self.uid = email if email_provider?
+    self.uid = email if uid_and_provider_defined? && email_provider?
```
MaicolBen, line 33: where is this defined?

MaicolBen (commented):
Can you explain your change?

app/models/devise_token_auth/concerns/user_omniauth_callbacks.rb
thooams: above line 21 ;)

thooams:
> Can you explain your change?
I edited my first comment.

thooams:
Someone wants to merge this PR ?

MaicolBen (approved):
@thooams Thanks, sorry for the delay, it's weird handling this case but it makes sense

PR #1520, opened 2022-01-01
add `previous_token`

### LINKED ISSUE #1232: Token invalidation after canceled request by the frontend app
I'm using devise token auth and implementing my own logic at the front-end to handle the token management.

After many of my users reporting a sudden token invalidation, I found out that this is caused when the frontend makes the request and cancels it (it happens when we refresh the browser in the middle of the promise runtime).

When the backend generates a new token but the frontend app is no longer listening to the response, the frontend only has the "last token" to work with instead of the newest one. After five seconds, the old token expires and the user is kicked out to the login page.

Any ideas to solve this?

Considering I could implement any logic in the frontend to put the authentication back on track after this event, I would still need more functionality in the devise token auth lib?

sudhanshu16:
Closes #1232

Solves the issue mentioned in #1232 

Implements the solution mentioned in https://github.com/lynndylanhurley/devise_token_auth/issues/1232#issuecomment-435022610
and basically copied from the PR #1315 

Added tests for the same, updated a few comments to separate keyword `previous` from `last`

(opened this PR as #1315  seems to be stale and this is a critical issue, at least for our use case)

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -120,6 +120,7 @@ module DeviseTokenAuth::Concerns::User
     # ghetto HashWithIndifferentAccess
     expiry     = tokens[client]['expiry'] || tokens[client][:expiry]
     token_hash = tokens[client]['token'] || tokens[client][:token]
+    previous_token_hash = tokens[client]['previous_token'] || tokens[client][:previous_token]
 
     return true if (
       # ensure that expiry and token are set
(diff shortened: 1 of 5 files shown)


MaicolBen (approved):
Great work! Thanks for the details and tests

sudhanshu16:
Thanks for the quick review @MaicolBen, any idea when this might be released?

Pifoupif:
@sudhanshu16 @MaicolBen  Was this ever merged ? I do not see it locally. How can we help making it part of the latest gem version ?

PR #1525, opened 2022-02-02
Fix for DeviseTokenAuth.change_headers_on_each_request being ignored

juddmon:
If DeviseTokenAuth.change_headers_on_each_request is set to false then the code falls into the else block and probably should not.


MaicolBen (commented):
can you add a test?

Issue #1456, opened 2021-01-17
rake db:migrate complains of multiple databases and an old OmniAuth version


mgmodell:
I am trying to run the tests on a clean pull of HEAD as of 1/17/2021 (c922580), but the bundle install takes hours and once it completes, it complains of a multiple database configuration and an old OmniAuth version (but it's using 2.0.1). If this is not an issue, please tell me what I'm doing wrong.

* **Version**: Head (as of 1/17)
* **Request and response headers**: n/a
* **Stacktrace**:
$ bundle exec rake db:migrate
Rails couldn't infer whether you are using multiple databases from your database.yml and can't generate the tasks for the non-primary databases. If you'd like to use this feature, please simplify your ERB.
rake aborted!
You are using an old OmniAuth version, please ensure you have 1.0.0.pr2 version or later installed.
/home/micah/dev/devise_token_auth/test/dummy/app/active_record/unregisterable_user.rb:5:in `<class:UnregisterableUser>'
/home/micah/dev/devise_token_auth/test/dummy/app/active_record/unregisterable_user.rb:4:in `<top (required)>'
/home/micah/dev/devise_token_auth/lib/devise_token_auth/rails/routes.rb:35:in `mount_devise_token_auth_for'
/home/micah/dev/devise_token_auth/test/dummy/config/routes.rb:17:in `block in <top (required)>'
/home/micah/dev/devise_token_auth/test/dummy/config/routes.rb:3:in `<top (required)>'
/home/micah/dev/devise_token_auth/test/dummy/config/environment.rb:7:in `<top (required)>'
/home/micah/.asdf/installs/ruby/2.7.2/bin/bundle:23:in `load'
/home/micah/.asdf/installs/ruby/2.7.2/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate => app:db:migrate => app:db:load_config => app:environment
(See full trace by running task with --trace)

* **Environmental Info**: asdf version manager with ruby 2.7.2

gamesover:
it seems can only support omniauth version < 2 for now

PR #1539, opened 2022-06-24
Set cookie token on omniauth success to avoid having to store the scraped query param token in the client

theblang:
In #1453 we added support for sending and receiving the auth token as a cookie. We're currently working on a de-angularized version of ng-token-auth with cookie support, which means we don't need to store the token in client storage anymore. But I realized that the OAuth flow still requires client storage because the token is scraped from query params, temporarily stored in client storage, then sent to the `validate_token` call, where a cookie is created. I realized that we could instead just create the cookie immediately in the same place that we're creating the query params in order to avoid having to leak the token in client storage just for that small initial period.

### DIFF
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -59,6 +59,14 @@ module DeviseTokenAuth
       set_token_on_resource
       create_auth_params
 
+      # If using a cookie to transport the token then we can set that cookie now, rather than making
+      # the client scrape the token from query params to then send in the initial validate_token request.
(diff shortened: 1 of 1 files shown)


theblang:
Hey @MaicolBen ! Quick question, I started to write a test for this, but when following the instructions I hit the following error:

```
Rails couldn't infer whether you are using multiple databases from your database.yml and can't generate the tasks for the non-primary databases. If you'd like to use this feature, please simplify your ERB.
```

I think this was also mentioned in #1456 . Any ideas of how I should proceed?

theblang:
Going to close this PR in favor of opening a new one whose source isn't my fork's master. Need to make a similar additional change for the reset password flow.

theblang:
See #1542 instead

Issue #1316, opened 2019-07-18
 Confirmations controller route error not found


igorgesso:
**rails (5.2.3)
devise_token_auth (1.1.0)
devise (4.6.2)**

I'm trying to make the confirmation part of the devise auth token work, but I'm running into a problem.

When accessing the link sent in the email always goes to the exception in the confirmations_controller as it is in the log and the route is not found.

I tried to understand the confirmations_controller to see what could be the problem, if anyone can give me some hint of what I might be doing wrong.

sorry for my english translate.

[Confirmation.log](https://github.com/lynndylanhurley/devise_token_auth/files/3407768/Confirmation.log)

igorgesso:
can someone explain how is the confirmable flow type what has in faq for reset password flow? You don't need as many details as faq.

experimatt:
@igorgesso did you ever figure out how to prevent this error? I'm having the exact same problem, and can't find good documentation on how the user email confirmation flow should work. Like which parameters are required (e.g. `email` and `redirect_url`), and which endpoint to hit (e.g. POST to `/auth/confirmation`).

It clearly works to _create_ the confirmation request (which sends an email), but upon clicking the link, I get the same NotFound error.

```
9:18:40 PM api.1 |  Started GET "/api/auth/confirmation?config=default&confirmation_token=eyY53XBy2sn1vg5DKC5u&redirect_url=http%3A%2F%2Flocalhost%3A3000%2Faccount" for ::1 at 2020-01-29 21:18:40 -0600
9:18:40 PM api.1 |  Processing by DeviseTokenAuth::ConfirmationsController#show as HTML
9:18:40 PM api.1 |    Parameters: {"config"=>"default", "confirmation_token"=>"eyY53XBy2sn1vg5DKC5u", "redirect_url"=>"http://localhost:3000/account"}
9:18:40 PM api.1 |  Unpermitted parameters: :config, :redirect_url
9:18:40 PM api.1 |  Completed 404 Not Found in 8ms (ActiveRecord: 1.2ms)
9:18:40 PM api.1 |  ActionController::RoutingError (Not Found):
9:18:40 PM api.1 |  devise_token_auth (1.1.3) app/controllers/devise_token_auth/confirmations_controller.rb:28:in `show'
```

experimatt:
I just figured out what caused this - I was trying to confirm an email that was already confirmed.

[...]

Issue #1123, opened 2018-03-22
Confirming an already confirmed user -- still not quite working.


Evan-M:
_**NOTE:** Much of this has been referenced before in several different Issues and PRs, which I will attempt to link to. If I've left any out, I apologize in advance._

-------

Although it appears that this was first noted in either #410 or #519, and eventually merged in #1001, the current behavior is still not great.

After confirming a user's email address, I can continue to click the confirmation link (i.e. make GET requests to the `/auth/confirmations` endpoint, triggering the `devise_token_auth/confirmations#show` action).  It seems that the current behavior will generate a _new_ auth token for the user, and redirect to the successful confirmation URL (either from the params if present, or the `DeviseTokenAuth.default_confirm_success_url`).  

As pointed out in #1064, #1067, and #1075, when the _new_ token is created, it conditionally adds a expiration if the user has signed in (the `:sign_in_count` attribute), which creates a dependency on `:trackable`. And while #1064 + #1061 effectively uncouple the dependency, the result is that if `:trackable` is not included, then no expiration is set on the token! What-the-huh!!?

The confirmation process _should_ happen in a consistent manner, regardless of the presence of the `:trackable` module.  A review of the Devise internals shows that checking if a user has signed in (i.e. `@resource.has_attribute?(:sign_in_count) && @resource.sign_in_count > 0`) is **not an indicator of the user having been confirmed!**  Rather, it is the presence of a timestamp in the `:confirmed_at` attribute that indicates if a user has confirmed their account. (See: [`Devise::Models::Confirmable#confirm`] and [`Devise::Models::Confirmable#confirmed?`])

Furthermore, the [`Devise::ConfirmationsController#show`] action simply and concisely checks for the presence of validation errors on the user resource, as it handles checking for already confirmed users at the model level. (The [`Devise::Models::Confirmable#confirm`] method sets errors on the user's `:email` attribute, if that user has already been confirmed).

We're already using the same [`Devise::Models::Confirmable::ClassMethods#confirm_by_token`] method in: https://github.com/lynndylanhurley/devise_token_auth/blob/38d27cfa80042c184c6f69e74c0789ca8ec0e407/app/controllers/devise_token_auth/confirmations_controller.rb#L3-L4

[...]

Issue #1110, opened 2018-03-14
Email confirmation route


eugenio1590:
This is a question, not an issue. In the documentation, I do not find the route to resending the email confirmation instructions. The default route for the Devise library is POST /confirmation with the email parameter (access is with /confirmation/new route). Which is the route in this library?

zachfeldman:
Hey @eugenio1590 , what happens when you run `rake routes` - do you see it then?

eugenio1590:
Hi @zachfeldman I can see the route: 

````
new_api_v1_user_confirmation GET      /api/v1/auth/confirmation/new(.:format)             devise_token_auth/confirmations#new
api_v1_user_confirmation     GET      /api/v1/auth/confirmation(.:format)                 devise_token_auth/confirmations#show
                            POST      /api/v1/auth/confirmation(.:format)                 devise_token_auth/confirmations#create

````

But, when I access the first link, I get the following error:
`The action 'new' could not be found for DeviseTokenAuth::ConfirmationsController`

And the second link: `Routing Error Not Found`

And the third link: `The action 'create' could not be found for DeviseTokenAuth::ConfirmationsController`

zachfeldman:
This is the ConfirmationsController on master:
https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/confirmations_controller.rb

It seems to be lacking some actions?

zachfeldman:
...or not, depending on which endpoint you're hitting.

You can also override that controller and do something like binding.pry to debug what your request sends.

eugenio1590:
Ok, I don't see the 'new' and 'create' actions in this controller. The problem is the following, this is my user case in my android application:

1. The user does not have an account.
2. The user presses the 'sign-up' button.
3. The user enters their credentials (email and password).
4. In the Rails server, the account is created and send a confirmation email to the user.
5. The app tries to login with these credential, but display an information message about the email confirmation.

---The functionality is right until here---

Sometimes (due to my mail provider used in the Rails server), the confirmation email does not reach. I need to implement a way to resend this confirmation email, like the Devise library.

zachfeldman:
Sure! I'd accept a PR for a route.

Evan-M:
@eugenio1590 funny, I noticed the exact same thing last week, but I've not had a chance to report the issue.

[...]

PR #1557, opened 2022-09-30
Avoid raising a RoutingError when confirming a user twice

micred:
A simple PR to avoid raising a RoutingError when confirming a user twice or when the token has expired.
Please check it out.

#1123
#1110
#1316


MaicolBen (commented):
can you add a spec?

PR #1534, opened 2022-03-23
Added 'Authorization' header with bearer token

rhiroshi:
Added Authorization header with bearer token defined in [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) 
related issues: #1522 #902 #1487

PR #1567, opened 2022-12-06
Remove bearer token if cookie_enabled is true

rhiroshi:
Some people were asking for the bearer token to not be included when using cookie_enabled: true in the [PR that added the bearer token](https://github.com/lynndylanhurley/devise_token_auth/pull/1534)


MaicolBen:
@rhiroshi can you fix the specs?

blackst0ne:
Any chance to get this merged?

MaicolBen:
@blackst0ne sorry, just checking this, can you rebase to make sure the build pass?

blackst0ne:
@MaicolBen 

This is not my PR. :)

MaicolBen:
Oh sorry, bad tag, @rhiroshi can you rebase?

rhiroshi:
@MaicolBen Rebased. Had some tests failing but was not related to my changes. I made another commit to trigger a rerun and the tests passed. So sorry for taking so long, had a heavy workload on the end of the last year and ended up forgetting to fix things here.

MaicolBen:
@rhiroshi no problem, thank you!

djpremier:
@MaicolBen Could we be reviewing a way to add the header? Because it is very useful in requests that are not via cookies, and it is the default reported by the Bearer token documentation. If you agree, I will think of a solution that does not honor cookies and both can coexist, since having the token explained in several "keys" is actually even more "expensive" in terms of size, which was the reason for removing it, because we could follow the pattern and the definition of the implementation and have only the "Authorization" header, instead of sending it separately.

What do you think? Removing the other headers would imply a break change, but we can do this in parts, starting with opening requests for changes to the libraries that use this lib.

djpremier:
@MaicolBen Could we be reviewing a way to keep the header even using cookies? Because using cookies does not eliminate the need for headers, since shell applications (and others) that do not support cookies need them, so there are applications (like the one I develop) that have both functionalities (authentication through cookies or headers).

As the main driver of this change is due to the size of the cookie, what do you think about following the pattern of other languages and even the documentation, returning only the "Authorization" header, that is, removing all others (`access-token` , `token-type`, `client`, `expiry`, `uid`) that we currently have, as they are "just" the "separate" form of the token, and in turn end up making the size bigger?

I understand that removing these other headers would imply a breaking change, but we can do this in parts, starting with a flag to return in this new format and opening pull requests for changes in libraries that use this lib (eg `ng-token -auth`, `jToker`, etc).

What do you think? Could we go down this path? If so, I can start crafting.

PR #1568, opened 2022-12-15
Fix unpermitted parameters warning

remy727:
### WHY are these changes introduced?

[Rails 7 provides context when logging unpermitted parameters](https://blog.saeloun.com/2021/06/16/rails-7-provides-context-when-logging-unpermitted-parameters.html)

### WHAT is this pull request doing?

Permit `session `parameter

### Screenshots
image

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -131,7 +131,7 @@ module DeviseTokenAuth
     private
 
     def resource_params
-      params.permit(*params_for_resource(:sign_in))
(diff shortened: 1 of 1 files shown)


MaicolBen:
@remy727 This broke a bunch of specs (I missed that the tests weren't run), can you open a new PR and fix them?

remy727:
@MaicolBen, yes for sure.
Do I need to create `.env` file in **root** directory or `test/dummy` folder?

MaicolBen:
@remy727 sorry, just saw this, you don't need an env file, just install (I think you need to setup one db, you can comment the rest in the gemspec) and run `rake test`

PR #1569, opened 2022-12-16
Fixed vulnerabilities

ryanfox1985:
- unsafe reflection
- sql injection

### DIFF
--- a/app/controllers/devise_token_auth/concerns/resource_finder.rb
+++ b/app/controllers/devise_token_auth/concerns/resource_finder.rb
@@ -22,7 +22,8 @@ module DeviseTokenAuth::Concerns::ResourceFinder
   def find_resource(field, value)
     @resource = if database_adapter&.include?('mysql')
                   # fix for mysql default case insensitivity
-                  resource_class.where("BINARY #{field} = ? AND provider= ?", value, provider).first
+                  field_sanitized = resource_class.connection.quote_column_name(field)
+                  resource_class.where("BINARY #{field_sanitized} = ? AND provider= ?", value, provider).first
                 else
                   resource_class.dta_find_by(field => value, 'provider' => provider)
                 end
--- a/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
@@ -133,13 +133,13 @@ module DeviseTokenAuth
     end
 
     def resource_class(mapping = nil)
-      if omniauth_params['resource_class']
-        omniauth_params['resource_class'].constantize
-      elsif params['resource_class']
-        params['resource_class'].constantize
-      else
-        raise 'No resource_class found'
-      end
+      return @resource_class if defined?(@resource_class)
+
+      constant_name = omniauth_params['resource_class'] || params['resource_class']
+      @resource_class = ObjectSpace.each_object(Class).detect { |cls| cls.name == constant_name }
+      raise 'No resource_class found' if @resource_class.nil?
+
+      @resource_class
     end
 
     def resource_name
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -176,10 +176,10 @@ module DeviseTokenAuth::Concerns::User
       updated_at: now
     )
 
-    update_auth_header(token.token, token.client)
+    update_auth_headers(token.token, token.client)
   end
 
-  def build_auth_header(token, client = 'default')
+  def build_auth_headers(token, client = 'default')
     # client may use expiry to prevent validation request if expired
     # must be cast as string or headers will break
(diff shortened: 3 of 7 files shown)


app/controllers/devise_token_auth/concerns/resource_finder.rb
```diff
@@ -22,7 +22,8 @@ def get_case_insensitive_field_from_resource_params(field)
   def find_resource(field, value)
     @resource = if database_adapter&.include?('mysql')
                   # fix for mysql default case insensitivity
-                  resource_class.where("BINARY #{field} = ? AND provider= ?", value, provider).first
+                  field_sanitized = resource_class.connection.quote_column_name(field)
+                  resource_class.where("BINARY #{field_sanitized} = ? AND provider= ?", value, provider).first
```
ryanfox1985, line 26:
```
Confidence: Weak
Category: SQL Injection
Check: SQL
Message: Possible SQL injection
Code: resource_class.where("BINARY #{field} = ? AND provider= ?", value, provider)
File: gems/devise_token_auth/app/controllers/devise_token_auth/concerns/resource_finder.rb
Line: 25
```

app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
```diff
@@ -133,13 +133,13 @@ def whitelisted_params
...
-      else
-        raise 'No resource_class found'
-      end
+      return @resource_class if defined?(@resource_class)
+
+      constant_name = omniauth_params['resource_class'] || params['resource_class']
+      @resource_class = ObjectSpace.each_object(Class).detect { |cls| cls.name == constant_name }
+      raise 'No resource_class found' if @resource_class.nil?
+
+      @resource_class
```
ryanfox1985, line 142:
```
Confidence: High
Category: Remote Code Execution
Check: UnsafeReflection
Message: Unsafe reflection method `constantize` called on parameter value
Code: params["resource_class"].constantize
File: gems/devise_token_auth/app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
Line: 135
```

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -190,7 +190,7 @@ def build_auth_header(token, client = 'default')
       DeviseTokenAuth.headers_names[:"expiry"]       => expiry.to_s,
       DeviseTokenAuth.headers_names[:"uid"]          => uid
     }
-    headers.merge!(build_bearer_token(headers))
+    headers.merge(build_bearer_token(headers))
```
ryanfox1985, line 193: useless bang.

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -176,10 +176,10 @@ def create_new_auth_token(client = nil)
       updated_at: now
     )
 
-    update_auth_header(token.token, token.client)
+    update_auth_headers(token.token, token.client)
```
ryanfox1985, line 179: I decided to pluralize the method.

MaicolBen:
@ryanfox1985 can you fix the specs?

app/controllers/devise_token_auth/omniauth_callbacks_controller.rb
MaicolBen: @ryanfox1985 I had to [rollback](https://github.com/lynndylanhurley/devise_token_auth/pull/1582) this because broke the build (it seems at the time that the CI's status wasn't showing correctly), if you know a different way to fix that avoids this unsafe execution, let me know.

ryanfox1985: no worries, this PR includes some refactors besides the vulnerability fixes let me reopen in three different PRs.

ryanfox1985: could you clarify what was broken in the build?

ryanfox1985: oh I see what happens, it's loaded test models with the same class name in the tests...

ryanfox1985: I found a work around to this, https://github.com/lynndylanhurley/devise_token_auth/pull/1587/files#diff-e297b1cbec8129624683adca1209383ebe6cbc3bfeebae01e590530e597dc062R138

basically the problem is in the tests some generated classes matches with name User for instance, in theory this patch covers that however, this only works for models under `ActiveRecord`... I don't think is possible, but, do you know if it's possible to have `User` without using ActiveRecord?

ryanfox1985: oh no, there is Mongoid...

MaicolBen: sorry, I missed these messages, glad you found the problem

PR #1570, opened 2022-12-22
Allow extra data for new token creation method

Zogoo:
`create_token` method is allowing extra data but, `create_new_auth_token` is blocking that feature.
And this will help users of this gem define their own logic with extra data.


MaicolBen:
@Zogoo can you fix the spec?

MaicolBen:
Re-opened in case I didn't run the specs, but yes, they are still failing

PR #1577, opened 2023-02-06
🐛 Not update cookies when is a batch request

djpremier:
Due to the batch requests approach, when this occurs the header has a blank token for the other requests and with that the current logic invalidates the cookie, as it is automatically updated by the back-end without any control from the client (browser) , for those cases due to ignore any change in them to keep what was received in the last valid request (in theory the first received that was not treated as a batch).


MaicolBen:
@djpremier can you rebase to make sure tests pass?

djpremier:
> @djpremier can you rebase to make sure tests pass?

Hi Maicol, yes, one a moment.

MaicolBen:
one spec fails 😞

djpremier:
Hi @MaicolBen, now all tests work! 🎉

(I suspect it was `sprockets` fault as it also prevented me from uploading a "clean" project)

PR #1578, opened 2023-02-10
chore: add vanilla-token-auth to client list

theblang:
We're in the process of converting AngularJS code to React and needed a client that could interface with Devise Token Auth in both contexts, so we built this vanilla client based on [ng-token-auth](https://github.com/lynndylanhurley/ng-token-auth). I thought it might be helpful to others, so I'm adding it to the client list in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Also, it maintains a session for each client/device, so you can have as many ses
 ## Main features
 
 * Seamless integration with:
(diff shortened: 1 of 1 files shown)


theblang:
Hey @MaicolBen! I'm a coworker of @booleanbetrayal . Last year we did a decaffeinate/de-angularize pass on `ng-token-auth` so that we could do auth from both AngularJS and React land. Was thinking it might be helpful to add it to the list, for anyone else that may be moving away from AngularJS. Let me know what you think!

PR #1599, opened 2023-05-23
add redirect_to options for rails7 allow_other_host

ihatov08:
In Rails 7, it's necessary to add the allow_other_host option when redirecting to an external path. This has also been made possible in Devise Token Auth, where the allow_other_host option can now be set in redirect_to.

https://api.rubyonrails.org/v7.0.4.2/classes/ActionController/Redirecting.html#method-i-redirect_to

> To allow any external redirects pass allow_other_host: true, though using a user-provided param in that case is unsafe.

```ruby
redirect_to "https://rubyonrails.org", allow_other_host: true
```

### DIFF
--- a/app/controllers/devise_token_auth/application_controller.rb
+++ b/app/controllers/devise_token_auth/application_controller.rb
@@ -84,6 +84,10 @@ module DeviseTokenAuth
       end
     end
 
+    def redirect_options
+      {}
+    end
+
     # When using a cookie to transport the auth token we can set it immediately in flows such as
     # reset password and OmniAuth success, rather than making the client scrape the token from
     # query params (to then send in the initial validate_token request).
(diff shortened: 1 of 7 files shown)


MaicolBen:
Weird, I don't think your changes break what is failing in the CI

edusurf10:
How do I pass these redirect options?

ihatov08:
@edusurf10 

Please override the redirect_options method.

example:

```ruby
# app/controllers/overrides/token_validations_controller.rb
module Overrides
  class TokenValidationsController < DeviseTokenAuth::TokenValidationsController
     ...

     protected
     
     def redirect_options
       { allow_other_host: true }
     end
  end

edusurf10:
@ihatov08 

But this method is defined in the ApplicationController, do I have to override it in this controller? Or do I need to override it in all the controllers of Devise Token Auth?
Example:
PasswordsController, etc...

PR #1600, opened 2023-05-31
Don't leak information about the existence of accounts in SessionsController

moritzhoeppner:
The SessionsController leaks information about the existence of accounts even if Devise's paranoid mode is activated. It leaks this information in the following ways:
1. If the resource is not found, the password isn't hashed, resulting in a probably significantly lower response time (given that the cost factor is configured correctly).
2. If an account is locked or not confirmed, the error messages say so.

The pull request fixes these issues by:
1. Hashing the given password if paranoid mode is activated and the resource isn't found.
2. Always responding with "bad credentials" if paranoid mode is activated and the login fails.

### DIFF
--- a/app/controllers/devise_token_auth/sessions_controller.rb
+++ b/app/controllers/devise_token_auth/sessions_controller.rb
@@ -34,13 +34,19 @@ module DeviseTokenAuth
         yield @resource if block_given?
 
         render_create_success
-      elsif @resource && !(!@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?)
+      elsif @resource && !Devise.paranoid && !(!@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?)
         if @resource.respond_to?(:locked_at) && @resource.locked_at
           render_create_error_account_locked
         else
           render_create_error_not_confirmed
         end
       else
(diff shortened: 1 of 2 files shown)


app/controllers/devise_token_auth/sessions_controller.rb
```diff
@@ -34,13 +34,19 @@ def create
...
         render_create_success
-      elsif @resource && !(!@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?)
+      elsif @resource && !Devise.paranoid && !(!@resource.respond_to?(:active_for_authentication?) || @resource.active_for_authentication?)
         if @resource.respond_to?(:locked_at) && @resource.locked_at
           render_create_error_account_locked
         else
           render_create_error_not_confirmed
         end
       else
+        # In order to avoid timing attacks in paranoid mode, we want the password hash to be
```
MaicolBen, line 44: can you move this new code and explanation to another method so doesn't make bigger than it's?

moritzhoeppner: Hi, thanks for the review :)
I moved the code in a new method and also took the liberty to refactor the beginning the create method a bit. Hope that's fine.

MaicolBen: thank you!

PR #1607, opened 2023-09-13
Faker safe_email -> email

hatsu38:
NOTE: Faker::Internet.safe_email is deprecated; use email instead. It will be removed on or after 2023-10.
FYI: https://github.com/faker-ruby/faker/blob/56a96cc3d54b9249f8062ebf1779deabc4a282f3/lib/faker/default/internet.rb#L87
```ruby
            deprecate :safe_email, :email, 2023, 10
```

The `safe_email` method is deprecated. The `email` method is safe by default.
https://github.com/faker-ruby/faker/pull/2733

This is a local test result.
スクリーンショット 2023-09-13 19 07 35


MaicolBen:
This broke some specs, I fixed it here https://github.com/lynndylanhurley/devise_token_auth/pull/1613

PR #1614, opened 2023-11-08
Fix Rails version comparison

santib:
the previous condition would return false for Rails 8.0, when what we actually want is to always return true for versions > 7.1, right?

### DIFF
--- a/app/models/devise_token_auth/concerns/active_record_support.rb
+++ b/app/models/devise_token_auth/concerns/active_record_support.rb
@@ -2,7 +2,7 @@ module DeviseTokenAuth::Concerns::ActiveRecordSupport
   extend ActiveSupport::Concern
 
   included do
-    if Rails::VERSION::MAJOR >= 7 && Rails::VERSION::MINOR >= 1
+    if Rails.gem_version >= Gem::Version.new("7.1.0.a")
       serialize :tokens, coder: DeviseTokenAuth::Concerns::TokensSerialization
     else
       serialize :tokens, DeviseTokenAuth::Concerns::TokensSerialization


MaicolBen (approved):
Right! Thank you santi !!

app/models/devise_token_auth/concerns/active_record_support.rb
```diff
@@ -2,7 +2,7 @@ module DeviseTokenAuth::Concerns::ActiveRecordSupport
   extend ActiveSupport::Concern
 
   included do
-    if Rails::VERSION::MAJOR >= 7 && Rails::VERSION::MINOR >= 1
+    if Gem::Version.new(Rails::VERSION::STRING) >= "7.1"
```
MaicolBen, line 5: This fails in ruby < 3.1 with `ArgumentError: comparison of Gem::Version with String failed` . Maybe try with `Rails.gem_version >= Gem::Version.new("7.1.x")` ?

santib: Interesting. There I pushed your suggestion with a minor change to account for pre-release versions

image

santib: Interesting. There I pushed your suggestion with a minor change to account for pre-release versions

image

scott-knight:
@MaicolBen are we getting a tagged release for this in RubyGems soon? The [graphql_devise](https://github.com/graphql-devise/graphql_devise) update for Rails 7.1 is on hold until there is an updated release from devise_token_auth. Thanks in advance.

scott-knight:
@MaicolBen hate to bug you about this, but this merge needs to be tagged for a RubyGems release. The [graphql_devise](https://github.com/graphql-devise/graphql_devise) update for Rails 7.1 is on hold until there is an updated release from devise_token_auth. Thanks in advance.

Issue #1618, opened 2024-02-29
Don't allow the newly created token to be ejected in `clean_old_tokens`


theblang:
We hit an odd situation with our Pingdom user, where it ended up having 10 tokens (the default `max_number_of_devices`) with an `expiry` slightly too far into the future. That in itself is still a mystery. But it caused a situation where the user couldn't log in, because the newly created token (with a proper `expiry`) was the one being ejected in `clean_old_tokens`. I think we should change that logic to never eject the token that was just created, otherwise the user can't sign in. Barring my odd situation, you might do something like change `token_lifespan` to be earlier.

dalima-dev:
Got same problem here, me and my team suspect of this year being a leap year is causing that behavior.

theblang:
> me and my team suspect of this year being a leap year is causing that behavior.

Ahh, that's a good thought!

pedroara:
Although I think we should focus on cause of the tokens that get generated with a larger expiry than the configured lifespan the changes you suggest could solve the problem

MaicolBen:
I could know how to fix it but I wasn't able to reproduce it with a test, so any more reproduction info is useful.

javaharut:
I also faced such weird issue during my work so solved it and created pr, hope it will fix it. 
https://github.com/lynndylanhurley/devise_token_auth/pull/1643

leightjohnson93:
I just opened PR that addresses this https://github.com/lynndylanhurley/devise_token_auth/pull/1657

PR #1643, opened 2024-12-31
Always ensure the newly created token expiry is not behind from exist…

javaharut:
Fix for https://github.com/lynndylanhurley/devise_token_auth/issues/1618

It ensures that the new token's expiration date won't be erased during the creation process since in certain cases, the old tokens' expiration date may fall after the newly formed one.

### DIFF
--- a/app/models/devise_token_auth/concerns/user.rb
+++ b/app/models/devise_token_auth/concerns/user.rb
@@ -92,9 +92,12 @@ module DeviseTokenAuth::Concerns::User
     def create_token(client: nil, lifespan: nil, cost: nil, **token_extras)
       token = DeviseTokenAuth::TokenFactory.create(client: client, lifespan: lifespan, cost: cost)
 
+      max_expiration_token = tokens.max_by { |_, token_info| token_info['expiry'] }
+      max_expiry = max_expiration_token&.dig(1, 'expiry') || 0
+
       tokens[token.client] = {
-        token:  token.token_hash,
-        expiry: token.expiry
+        token: token.token_hash,
+        expiry: [token.expiry, max_expiry + 1].max
       }.merge!(token_extras)
 
       clean_old_tokens


MaicolBen:
can you add a test?

app/models/devise_token_auth/concerns/user.rb
```diff
@@ -92,9 +92,12 @@ def send_unlock_instructions(opts = {})
...
       token = DeviseTokenAuth::TokenFactory.create(client: client, lifespan: lifespan, cost: cost)
 
+      max_expiration_token = tokens.max_by { |_, token_info| token_info['expiry'] }
+      max_expiry = max_expiration_token&.dig(1, 'expiry') || 0
+
       tokens[token.client] = {
-        token:  token.token_hash,
-        expiry: token.expiry
+        token: token.token_hash,
+        expiry: [token.expiry, max_expiry + 1].max
```
MaicolBen, line 100: why `+ 1` ?

javaharut: `max_expiry` is a timestamp and the expiry for the new token should be bigger than the current one and 1 is enough to achieve the goal

MaicolBen: sounds good, can you add a test?

javaharut: @MaicolBen I tried to add but unfortunately cannot run test in my local. there is an instruction error in low level. How can I get help with this?

MaicolBen: That isn't enough details, anyway if you get an idea on how write it, share it here and I can make it work.

MaicolBen:
I believe https://github.com/lynndylanhurley/devise_token_auth/pull/1657 fixed the same issue, let me know otherwise

PR #1650, opened 2025-03-01
Include reset_password_token in the redirect URL for the password reset flow

kaekasui:
This PR addresses an issue where the password reset email link includes `reset_password_token`, but the token is lost in the subsequent redirect URL. Since the final step of resetting the password requires `reset_password_token` as a parameter, users were unable to complete the password reset unless the token was explicitly included in the final URL.

By adding `reset_password_token` to redirect_header_options, the token is preserved in the redirect URL, allowing the user to properly finish the password reset flow.

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -56,7 +56,7 @@ module DeviseTokenAuth
             set_token_in_cookie(@resource, token)
           end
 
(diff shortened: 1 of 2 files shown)


MaicolBen (commented):
Can you rebase because the test suite was fixed and add a test?

kaekasui:
Rebased with the latest changes :)

MaicolBen (approved):

PR #1662, opened 2025-08-07
Handle expired reset_password_tokens on passwords#update

catks:
Currently, `DeviseTokenAuth` uses `resource#with_reset_password_token(resource_params[:reset_password_token])` to retrieve the user in the `password#update`, which returns the user even if the token has expired.

The default Devise Controller uses [reset_password_by_token ](https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/app/controllers/devise/passwords_controller.rb#L34C21-L34C59), which returns an empty User with an error.

To keep the same flow, this PR adds another call to `resource#reset_password_period_valid?` to ensure the password is still valid.

Possible issues in real scenarios:

- A User opens the edit link but takes too long to submit
- A request is made directly to the update endpoint using the token from the edit link

### DIFF
--- a/app/controllers/devise_token_auth/passwords_controller.rb
+++ b/app/controllers/devise_token_auth/passwords_controller.rb
@@ -73,7 +73,7 @@ module DeviseTokenAuth
       # make sure user is authorized
       if require_client_password_reset_token? && resource_params[:reset_password_token]
         @resource = resource_class.with_reset_password_token(resource_params[:reset_password_token])
-        return render_update_error_unauthorized if @resource.blank? || (@resource && !@resource.reset_password_period_valid?)
+        return render_update_error_unauthorized unless @resource && @resource.reset_password_period_valid?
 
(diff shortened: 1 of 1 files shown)


app/controllers/devise_token_auth/passwords_controller.rb
```diff
@@ -73,7 +73,7 @@ def update
       # make sure user is authorized
       if require_client_password_reset_token? && resource_params[:reset_password_token]
         @resource = resource_class.with_reset_password_token(resource_params[:reset_password_token])
-        return render_update_error_unauthorized unless @resource
+        return render_update_error_unauthorized if @resource.blank? || (@resource && !@resource.reset_password_period_valid?)
```
aitorlb, line 76:
```suggestion
        return render_update_error_unauthorized unless @resource && @resource.reset_password_period_valid?
```

